PR #2, opened 2019-06-18
Fix register example key

oliverlin:
Fix incorrect example key under register section.

螢幕快照 2019-06-18 上午10 32 31


bluebill1049:
Thank you very much :)

PR #7, opened 2019-07-28
Fix slashes in regex validation example

chrisparton1991:
Just a quick fix, noticed this whilst looking at the validation docs.


bluebill1049:
ha, thanks @chrisparton1991 for your contribution on the website as well!

PR #8, opened 2019-08-07
fix(example): fix bug in defaultExample

tknickman:
Great library/site!

I was playing around with the examples and noticed there is a small issue with the code for the default example (first one here: https://react-hook-form.com/get-started). 

The two issues are:
1. The error object is checking the wrong field for errors (checking errors from the `example` field, not the `exampleRequired` field)
2. The error `span`, is being rendered a string instead of html. 

This fixes both of those issues. 

Thanks!


tknickman:
Here is an updated codeSandbox for the example as well: https://codesandbox.io/embed/react-hook-form-get-started-m07wd (with some minor additional css to highlight the validation text)

bluebill1049:
so good! thank you very much. ❤️

PR #12, opened 2019-08-27
remove duplicate handleSubmit wrapper

ManAnRuck:
the onSubmit function was wrapped two with`handleSubmit`  times in the typescript example


bluebill1049:
aha. thank you ❤️

PR #19, opened 2019-10-11
Fix typo in useForm register API docs

jbranchaud:
My understanding is that this was supposed to read, "... create nested _form_ fields".

If it is supposed to read a different way, I'm happy to update the PR.


bluebill1049:
Thank you very much! Oops that was my bad 🤗

PR #21, opened 2019-10-13
Add `clearError` to the side menu.

JeromeDeLeon:
I was about to add description for `validateSchemaOption` but never understood its uses like I thought setting `abortEarly` to `false` would return all errors instead of one.

### DIFF
--- a/src/components/ApiPage.tsx
+++ b/src/components/ApiPage.tsx
@@ -167,6 +167,7 @@ const links = [
   "handleSubmit",
   "reset",
   "setError",
+  "clearError",
   "setValue",
   "getValues",
   "triggerValidation",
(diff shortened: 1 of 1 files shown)


bluebill1049:
> I was about to add description for `validateSchemaOption` but never understood its uses like I thought setting `abortEarly` to `false` would return all errors instead of one.

`abortEarly` to `false`  return all errors which is default behavior

JeromeDeLeon:
> > I was about to add description for `validateSchemaOption` but never understood its uses like I thought setting `abortEarly` to `false` would return all errors instead of one.
> 
> `abortEarly` to `false` return all errors which is default behavior

Here's the [Codesandbox](https://codesandbox.io/embed/young-http-wyh2s). It's only returning either `max` or `matches` error.

bluebill1049:
@JeromeDeLeon I meant it will return all errors for the fields (first error of each)

JeromeDeLeon:
Now I get it. Thanks. I thought the other way around 🤣. Silly me.

bluebill1049:
nah all good @JeromeDeLeon :)

PR #28, opened 2019-10-18
Minor correction to add errors to useForm initialization

yusinto:
The `errors` object is missing from the destructuring example in the TypeScript section.


bluebill1049:
thanks @yusinto finally yay!

PR #31, opened 2019-10-20
translate home data into japanese

kotarella1110:


bluebill1049:
thanks you @kotarella1110 !

PR #44, opened 2019-10-26
fix typos in example codes

adhrinae:
I found some typos translating 'Adavanced' page

* in the accessbility example, `id` doesn't match with `aria-describedby`
* Wizzard -> Wizard


bluebill1049:
ha thank you very much! my bad :) @adhrinae

PR #62, opened 2019-11-03
Improve i18n navigations

adhrinae:
1. Applied `translatedLink` to some navigations which are missing it.
2. When changing the language with the `select` component on the upper right side, it also changes the current pathname following the selected language.

The second one looks a bit dirty and might be unreadable. We can revert it anytime.


bluebill1049:
This is truly amazing! thanks, @adhrinae I love how you fixed the select for languages

PR #72, opened 2019-11-06
Fix antd input change handler

LarsJK:


bluebill1049:
wow thank you :) my bad...

PR #78, opened 2019-11-12
improve get started page

bluebill1049:


src/data/getStarted.tsx
```diff
@@ -391,16 +392,17 @@ export default {
             component libraries.
           </p>
           <p>
-            <span style={{ color: colors.lightPink }}>Note:</span> Most UI
-            libraries do expose <code>innerRef</code> or <code>ref</code> to{" "}
+            <Note>Note:</Note> Most UI libraries do expose <code>innerRef</code>{" "}
```
bluebill1049: hi @adhrinae @kotarella1110 ,

I have updated this section to make thing bit more clear:

> Note: Most UI libraries do expose innerRef or ref to register. However, if that's not the case. For example react-select or react-datepicker, you can use register at useEffect and update the value via setValue or trigger an error with setError.

please update the translation when you have sometimes :) 

cheers
bill

bluebill1049: no worries at all :) thank you

bluebill1049:
going to merge this one in first, you guys can follow it up whenever you are free :) thank you

PR #86, opened 2019-11-18
fix validation example on a main page

dchekmarev:
validation function should returns true on success and error string on failure.
here's example of original behaviour and fixed:
image


bluebill1049:
thanks @dchekmarev

PR #90, opened 2019-11-20
Fix language in API documentation

Betree:
I'm not a native English speaker, but the following looks more correct to me.


bluebill1049:
thank you and you are right :)

PR #97, opened 2019-11-26
update errors description and types in api page

kotarella1110:


src/data/api.tsx
```diff
@@ -1192,6 +1200,14 @@ export default {
           </p>
         </>
       ),
```
kotarella1110: @adhrinae  please help translate this when you have some free time!

src/data/api.tsx
```diff
@@ -1266,6 +1289,14 @@ export default {
...
```
kotarella1110: @bluebill1049  please help translate this when you have some free time!

src/data/api.tsx
bluebill1049: Thanks I will do that tonight

PR #98, opened 2019-11-26
fix RHFInput description in api page

kotarella1110:


src/data/api.tsx
```diff
@@ -2319,92 +2319,188 @@ export default {
...
-            you all of the <code>useForm</code> hook functions.
+            You can read more at{" "}
+            <a
+              href="https://github.com/react-hook-form/react-hook-form-input"
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              react-hook-form-input repo
+            </a>
+            .
```
kotarella1110: @adhrinae  please help translate this when you have some free time!

src/data/api.tsx
```diff
@@ -2319,92 +2319,188 @@ export default {
...
```
kotarella1110: @bluebill1049   please help translate this when you have some free time!

bluebill1049: thanks @kotarella1110 will do :)

PR #109, opened 2019-12-14
Update getStarted.tsx

cargallo:
Material UI TextField accepts inputRef asi reference not innerRef. I changed in all languages except in japanesse and chinese because they already differ from other languages translation.


bluebill1049:
thank you @cargallo <3

PR #119, opened 2019-12-24
Import control from useForm()

raikusy:
The `control` was missing. `register` and `setValue` was destructured for no use. Removed these two, and added control.


bluebill1049:
thank you @raikusy 🙏

PR #128, opened 2020-01-01
fix: broken link to advanced usage

mutewinter:
Needed to be absolute.

Note: this may also be an issue [on this line](https://github.com/react-hook-form/react-hook-form-website/blob/d8912d9521462d6fb2f04effb238110ae3c63e7e/src/data/es/api.tsx#L302), but I am unfamiliar with the codebase and unable to reproduce the error on that page.


bluebill1049:
ty very much :)

PR #129, opened 2020-01-02
Fix connect form example

catnose99:
Corrected `props.children` to `children` in connectForm example.
Thank you! 😸


bluebill1049:
oops thank you ❤️

PR #140, opened 2020-01-12
Fix invalid documentation mentioning error message in value property

LorenzHenk:


bluebill1049:
aha what a mistake. thanks

PR #150, opened 2020-01-19
update useFieldArray section in api jp data

kotarella1110:


bluebill1049:
i guess this is outdated :)

PR #158, opened 2020-01-28
Fix navigation error when using translated page

30in2020:
Hello.
When a user clicks the `advancedUsage` button on a translated page, 
the page doesn't show up correctly.
So I fixed a `link` parameter, from `/advanced-usage` to `advanced-usage`.


bluebill1049:
thank you :)

PR #161, opened 2020-02-02
remove styled component

bluebill1049:


adhrinae:
I wish `upgrade gatsby` commit could be separated from styling which means It should be just a commit including `package.json` & `yarn.lock` files.

and what's the context of removing styled-components? Is it about performance?

bluebill1049:
> I wish `upgrade gatsby` commit could be separated from styling which means It should be just a commit including `package.json` & `yarn.lock` files.
> 
> and what's the context of removing styled-components? Is it about performance?

yes @adhrinae performance :)

Issue #165, opened 2020-02-04
Incorrect Redux Form example in Library Code Comparison


chrissantamaria:
Hi all,

I'm not super familiar with Redux Form, but this code snippet from the Library Code Comparison section of the homepage caught my attention:

https://github.com/react-hook-form/react-hook-form-website/blob/eb3b125d5f91818419aebfcef4a7d669fe3cf3a9/src/components/codeExamples/reduxFormCode.ts#L9-L13

Shouldn't this just be checking `values.username`, not `values.username.length`? Happy to make PR if that's the case - just wanted to check first.

bluebill1049:
yes, your right please send a PR. 🙏

chrissantamaria:
Done! #166

bluebill1049:
thank you for your contribution @chrisparton1991

chrissantamaria:
You mean @chrissantamaria? :stuck_out_tongue:

bluebill1049:
Oops 😐

chrisparton1991:
You're welcome Bill, always happy to help 😂

bluebill1049:
LOL 👍

PR #166, opened 2020-02-04
Fix Redux Form example in Library Code Comparison

chrissantamaria:
This addresses #165. I've also fixed the equality check on the same line - unless I'm misinterpreting the goal of that validation, I believe it should be `values.username === "admin"` rather than `values.username !== "admin"`, similar to the Formik example.


bluebill1049:
thank you ❤️

PR #167, opened 2020-02-05
Fix prop name in Adapting existing form code example

arku:
`register={register}` -> `ref={register}`


bluebill1049:
Oops thank you 🙏

PR #168, opened 2020-02-05
Fix Advanced Usage broken link

arku:
The **Advanced Usage** button on https://react-hook-form.com/api takes to https://react-hook-form.com/api/advanced-usage which does not exist.

This PR fixes the broken link.


bluebill1049:
Thanks @arku again.

PR #169, opened 2020-02-05
Fixing reset button data

osmar2008:
Spreading default values to avoid material-ui error being printed on console.

> Material-UI: you have provided an out-of-range value `undefined` for the select component.
> Consider providing a value that matches one of the available options or ''.
> The available values are `10`, `20`, `30`.


bluebill1049:
Thank you mate :)

PR #172, opened 2020-02-06
Update api.tsx in spanish translation

cargallo:
Hi @bluebill1049 I updated Fields Arrays translation that you asked for few weeks ago. I was on vacations! Thanks for this great library. Keep moving foward, regggards!


bluebill1049:
thank you very much 🙏

PR #176, opened 2020-02-08
Multiple typo fixes and rephrases.

gilbsgilbs:
**Disclaimer:** I'm not a native english speaker, nor a user of this library.

I saw some typos and sentences I had trouble to understand on this website. Therefore I spent some time to try to improve it. I think it's better, but probably still very far from perfect since I'm neither a native english speaker nor a very good writer. I just hope that I haven't made more mistakes than there previously was :crossed_fingers: . Also, I did not review the entire website because I lost patience after a while.

Lastly, I wonder why the website is that slow. In particular, the [API](https://react-hook-form.com/api#register) page makes my CPU crazy. Changing the `Register options` radio buttons makes my Firefox lag for one whole second for no apparent reason. It looks slightly better on Chrome, but far from good actually. Overall, it gives me a bad first impression on this library (that claims to be so much faster than Formik and others).

### DIFF
--- a/src/data/en/api.tsx
+++ b/src/data/en/api.tsx
@@ -28,12 +28,11 @@ export default {
     validateCriteriaMode: (
       <>
         <p>
-          The default behavior <code>firstErrorDetected</code> will validate all
-          fields' validation and gather all the first error encounter.
+          When set to <code>firstErrorDetected</code> (default), only first error from
+          each field will be gathered.
         </p>
         <p>
-          With config set to <code>all</code>, all fields' validation will be
-          run through and gather all the errors encounter.
+          When set to <code>all</code>, all errors from each field will be gathered.
         </p>
       </>
     ),
@@ -52,8 +51,8 @@ export default {
     validateOnChange: (
       <>
         Validation will trigger on the <code>change</code> event with each
-        input, and lead to multiple re-renders. Not recommended: Consider this
-        as a bad performance practice.
+        input, and lead to multiple re-renders. Warning: the often comes with
+        a significant impact on performances.
       </>
     ),
     defaultValues: goToSection => (
@@ -107,7 +106,7 @@ export default {
     ),
     reValidateMode: (
       <p>
-        This option allows you to configure when inputs with errors getd
+        This option allows you to configure when inputs with errors get
         re-validated (by default, validation is triggered during an input
         change.) <Popup />
       </p>
@@ -122,8 +121,8 @@ export default {
     submitFocusError: (
       <>
         <p>
-          By default when the user submits a form and that contains an error,
-          the first field with an error will be focused.
+          When set to true (default) and the user submits a form that fails the
+          validation, set focus on the first field with an error.
         </p>
 
         <p>
@@ -161,9 +160,9 @@ export default {
       <>
         <p>
           This method will allow you to <code>unregister</code> a single input
-          or an array of inputs. This is useful when you register your input
-          during <code>useEffect</code> as custom register and to unregister it
-          after component unmount.
+          or an array of inputs. This is useful when you used a custom register
+          in <code>useEffect</code> and want to unregister it when the component
+          unmounts.
         </p>
         <p>
           <b className={typographyStyles.note}>Note:</b> When you unregister an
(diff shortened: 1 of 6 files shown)


bluebill1049:
thanks for the fixes and feedbacks ❤️ i will take a look at the API page performance today.

src/data/en/api.tsx
```diff
@@ -122,8 +121,8 @@ export default {
     submitFocusError: (
       <>
         <p>
-          By default when the user submits a form and that contains an error,
-          the first field with an error will be focused.
+          When set to true (default) and the user submits a form that fails the
+          validation, set focus on the first field with an error.
```
bluebill1049:
```suggestion
          validation, it will set focus on the first field with an error.
```

src/data/en/api.tsx
```diff
@@ -934,49 +928,49 @@ export default {
...
-        <code>props</code> becomes painful.
-      </p>
-    ),
-    description: (
       <>
         <p>
-          Once your form is wrapped with <code>FormContext</code>, the{" "}
-          <code>useFormContext</code>:{" "}
-          <code className={typographyStyles.typeText}>function</code> can be
-          invoked in its child component.
```
bluebill1049: I think this paragraph was missed out

bluebill1049:
sorry about my English is driving you losing your patient :( I guess the lib itself does need some quality uplift in terms of English, so please help us and others when you have some more free time and PR is always welcome.

bluebill1049:
> Lastly, I wonder why the website is that slow. In particular, the API page makes my CPU crazy. Changing the Register options radio buttons makes my Firefox lag for one whole second for no apparent reason. It looks slightly better on Chrome, but far from good actually. Overall, it gives me a bad first impression on this library (that claims to be so much faster than Formik and others).

can you provide us with a video record of such behavior?

gilbsgilbs:
Sorry for the prompt review and merge :zap: .

> sorry about my English is driving you losing your patient :(

I'm sorry, I didn't mean to be rude at all. After reading my initial comment, I can see how it could be interpreted in an offending way and I'm deeply sorry about that. That was unfair and disrespectful on my part, especially regarding the amazing work you and other contributors have already achieved on this documentation and on the library itself. Moreover, my english is far from irreproachable, so really don't take my words as a judgement or anything like that :pray: .

I just lost patience because it's a very repetitive work and I ended up being tired by it :sweat_smile: .

> I guess the lib itself does need some quality uplift in terms of English, so please help us and others when you have some more free time and PR is always welcome.

Yeah, it could always be improved, but most importantly, I think it does the job. I may contribute more should I replace Formik with React Hook Form, but I'm undecided at this point. Formik performances are awful in my ~20 inputs form for some reason, but I find kind of cumbersome having to pass `inputRef` to all the inputs of the form (I use Material UI). And I'm not 100% convinced it will improve the performances that much yet.

> can you provide us with a video record of such behavior?

I can try.

bluebill1049:
Thanks @gilbsgilbs for your help and understanding. 

In terms of performance, I can answer this one :) 

you don't need to pass innerRef instead Controller. the problem that I try to solve here is constant re-rendering, reconciler is cheaper but it becomes not so cheaper when your component trees grow.  It's up to everyone which tool to use, personally i encountered an issue with it and provided with my own solution and share amongst the community. i think it's worth to give a try rather than just read what's on the doc. 👍

gilbsgilbs:
Apparently, the API page is way faster in private browsing mode, so it must be one of my extensions messing things up somewhere. However, the radio button remains very slow in private browsing mode: https://drive.google.com/file/d/16Z5AdlqTbikv_72mKcgoFM4yPZDKw6z3/view . Do you want me to open a specific issue for this?

> you don't need to pass innerRef instead Controller.

Do you mean that wrapping all my inputs like this is acceptable in term of performances?

```jsx
import TextInput from '@material-ui/core/TextInput';

export function TextInput(props) {
  return <Controller as={TextInput} {...props} />
}
```

Combined with the `FormContext` provider, it would provide a very similar API to what Formik currently does for me (which is very close to raw Material-UI forms actually). If so, you're right that testing out React Hook Form wouldn't take me as much time as I initially expected and I definitely might give it a try :thinking: . However, to be fair with Formik, I would have to enable the `onChange` mode in React Hook Form, isn't it?

bluebill1049:
> Apparently, the API page is way faster in private browsing mode, so it must be one of my extensions messing things up somewhere. However, the radio button remains very slow in private browsing mode: https://drive.google.com/file/d/16Z5AdlqTbikv_72mKcgoFM4yPZDKw6z3/view . Do you want me to open a specific issue for this?

Found issue, it's due to syntax highlight, I will find a way to optimize that. yes :) issue would be good.

> However, to be fair with Formik, I would have to enable the onChange mode in React Hook Form, isn't it?

It's up to you in terms of when to trigger validation, you don't need to enable `onChagne` mode. The controller will take care of the validation model. 

however, even with onChange mode, there is still a difference. Even with `onChange` we only trigger when the error appeared and disappeared, no extra re-render while user typing the same valid or invalid input.

gilbsgilbs:
> It's up to you in terms of when to trigger validation, you don't need to enable onChagne mode. The controller will take care of the validation mode.

Oh 🤯. That makes a lot of sense to me now. Thank you. I really like the approach after all.

PR #179, opened 2020-02-12
Remove state machine hook import in connect form example

jimmy-liao-gogoro:
In Advanced Connect Form example, `useStateMachine` hook is not used and unnecessary.  
URL: https://react-hook-form.com/advanced-usage#ConnectForm


bluebill1049:
thank you mate :)

PR #182, opened 2020-02-18
translate isolateRender section in home into jp

kotarella1110:


bluebill1049:
thanks @kotarella1110

PR #190, opened 2020-02-23
Add Typing of CSS Modules

### LINKED ISSUE #171: Set type definition file for CSS Modules
In order to make Typescript compiler understand CSS Modules files, We need to generate `.d.ts` file in the project so that the compiler won't complain about that.

The simplest solution is creating an `global.d.ts` or `import.d.ts` file and write like this.

```ts
declare module '*.css' {
  interface IClassNames {
    [className: string]: string
  }
  const classNames: IClassNames;
  export = classNames;
}
```

but this is not enough. because it's so simple and you won't get the autocompletion from this typing.
[There's a way to improve this.](https://github.com/Quramy/typed-css-modules)

adhrinae:
Resolves #171 

* Introduced a new plugin `gatsby-plugin-css-modules-typings` in order to generate type definition files.
  - It's a plugin for Gatsby, so that we don't need to generate new a `d.ts` files every time we create a new CSS Module file. It will be generated when you fire up the development server.
* Added type definition files of CSS Modules.


bluebill1049:
wow, that's awesome. thanks a lot for the hard work! <3

PR #194, opened 2020-02-24
Update faq.tsx

brigolini:
reviews faq.js for portuguese.


bluebill1049:
Thank you very much, because we don't have a dedicated translator some of the pages I did translate via google translate... really appreciated your help.

PR #195, opened 2020-02-24
fix: redux form example

orteth01:
I was looking at the examples on the home page today and the end of the redux form example didn't look quite right. I'm pretty sure this is what it should be.


bluebill1049:
aha thank you

PR #211, opened 2020-03-03
Make 'as' prop in Controller examples consistent

orteth01:
in `RHFInput.ts`, it [specifies a preffered syntax fo the Controller component](https://github.com/react-hook-form/react-hook-form-website/blob/master/src/components/codeExamples/RHFInput.ts#L17-L18). This PR updates all controller examples to match the preferred syntax.

it also updates all instances of `RHFInput` to `Controller`.


src/components/codeExamples/RHFInput.ts
```diff
@@ -30,6 +30,7 @@ function App() {
         defaultValue={{ value: 'chocolate' }}
       />
       
+      {* // what is this example showing? can we remove it? *}
```
orteth01: don't merge w/ this here. i just wanted to flag the `<Controller />` below. i'm not really sure what it is there for.

bluebill1049:
awesome stuff, i will take a look during lunchtime ❤️ much appreciated.

src/components/codeExamples/RHFInput.ts
bluebill1049: This file is showing under Controller section, if u search the file usage. I should rename the file to be Controller

orteth01: I can rename the file 👍

Can i remove this last `Controller` in the example (lines 34-39)? They don't show an example of anything new, i think

src/components/codeExamples/reactNativeController.ts
```diff
@@ -1,27 +1,34 @@
-export default `import React from 'react';
-import { View, TextInput, Button } from 'react-native';
-import { useForm, Controller } from 'react-hook-form';
+export default `import React from "react";
```
orteth01: both `reactNativeController` and `reactNativeRHFInput` existed, but they were essentially showing the same thing. i decided to use `reactNativeRHFInput`.

so the old `reactNativeController` is *deleted*
and then i renamed `reactNativeRHFInput` to `reactNativeController`.

sorry if this is unnecessarily confusing

bluebill1049: Oh sweet awesome.

src/components/codeExamples/RHFInput.ts
bluebill1049: Yea sure

PR #223, opened 2020-03-18
Update Doc Controller example

fxamauri:
Update example in Controller when used in React native, message prop type invalid prop value, in react-native TextInput value returned by onChange must be a string.
This issue was raised in https://github.com/react-hook-form/react-hook-form/issues/769


bluebill1049:
oh man thank you so much, i have a ticket for myself to update this. you helped me <3.

PR #226, opened 2020-03-22
Controller:onChange callback have to return value instead of object

kirill3333:
Here (https://react-hook-form.com/api#Controller) we have an example for Controller with onChange callback.

As far as I see from release notes for V5.0.0 (https://github.com/react-hook-form/react-hook-form/releases/tag/v5.0.0) we need to return value instead of an object like `{ value: selected }`

I use react-hook-form with `https://ant.design/` and I've found it useful to update this example


bluebill1049:
thank you

PR #232, opened 2020-03-31
example for working with a virtualized list

311chaos:
This is my first time contributing to this project, so let me know what steps I might be missing.

The goal of this PR is to add an example on how RHF could work with a virtualized list (I was using react-window by Brian Vaughn).

I would guess that if accepted, the CodeSandBox example would want to be forked to NOT be under my account.

I have a CodeSandBox showing the issue that I started with:
https://codesandbox.io/s/virtulized-react-hook-form-issue-efynk

and one that has a work around:
https://codesandbox.io/s/virtulized-react-hook-form-fix-o75u2

as well as a simplified workaround without any styles and in a single file:
https://codesandbox.io/s/virtulized-react-hook-form-fix-simple-w38ty

### DIFF
--- a/src/components/AdvancedPage.tsx
+++ b/src/components/AdvancedPage.tsx
@@ -36,6 +36,7 @@ const enLinks = [
   advancedEn.conditionalControlledComponent,
   advancedEn.controlledMixedWithUnControlled,
   advancedEn.customHookWithValidationResolver,
+  advancedEn.workingWithVirtualizedList,
 ]
 
 function Advanced({ defaultLang }: { defaultLang: string }) {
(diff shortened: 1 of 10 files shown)


311chaos:
Hrm, ya, I seem to be missing something. I see the item in the left Nav on the FAQ page, but it is not appearing in the list.

bluebill1049:
@311chaos don't worry i will contribute and help to this, and i think we should move this into advanced as well. you happy for me to move?

311chaos:
Yes, I was uncertain on where it should go. Ha. Where you think it is most appropriate works for me.

bluebill1049:
awesome stuff! thanks very much for your contribution!

311chaos:
Thanks, I see what I missed now. Entries in FaqPage.tsx.

PR #235, opened 2020-04-02
Improving pt translation

vitorboccio:


bluebill1049:
Thank you very much @vitorboccio 🙏

PR #237, opened 2020-04-06
FIX: register call in useFieldArray examples

kbakba:
If provide `register` without calling, examples works incorectly.

https://codesandbox.io/s/nervous-raman-7en3o


bluebill1049:
ty @kbakba 🙏

PR #240, opened 2020-04-14
Added `defaultValue` to useFieldArray example

lucasconstantino:
The current docs as misleading when they should `append({ name: "test" })` as if the initial value of the input would be `"test"`, while if no `defaultValue` is added to the input, the newly appended input will be empty by default.


bluebill1049:
oops! i miss that. thank you.

PR #251, opened 2020-05-01
Hide scroll to top button when Form Builder is open

irekrog:
When Form Builder is opened we can click button "scroll to top". After clicked "scroll to top" "close button" from a Form Builder overlap a language selector. In this PR has been changed z-index Form Builder "modal" which is higher than "scroll to top" button.


bluebill1049:
thanks @irekrog 👍

PR #253, opened 2020-05-03
Insert 'clearError' method

eatdesignlove:
I've found 'clearError' missed in the example code and i inserted it.


bluebill1049:
aha thank you 🙏

PR #254, opened 2020-05-05
Fix api page, type miss match on UseFormOptions.validateCriteriaMode

daisuke-awaji:
Types of `userForm` argument described in api page does not match `UserFormOptions.validateCriteriaMode` in types.tsx(api v5).

https://github.com/react-hook-form/react-hook-form/blob/master/src/types.ts#L108

```tsx
export type UseFormOptions<
  FormValues extends FieldValues = FieldValues,
  ValidationContext extends object = object
> = Partial<{
  mode: Mode;
  reValidateMode: Mode;
  defaultValues: DeepPartial<FormValues>;
  validationSchema: any;
  validationResolver: ValidationResolver<FormValues, ValidationContext>;
  validationContext: ValidationContext;
  submitFocusError: boolean;
  validateCriteriaMode: 'firstError' | 'all'; // <-- 😱
}>;
```


bluebill1049:
thank you very much! you are right.

PR #259, opened 2020-05-09
fix validation schema go to section

irekrog:
* Fix go to section validationSchema

P.S.
I have noticed @bluebill1049 you have upgraded packages (also Prettier) so https://prettier.io/docs/en/options.html#arrow-function-parentheses
> First available in v1.9.0, default value changed from avoid to always in v2.0.0

and thats why Prettier in precommit reformatted these files


bluebill1049:
👍 thanks @irekrog again ❤️

PR #270, opened 2020-05-24
Add focus as a ref usage case

dandv:
Per https://github.com/react-hook-form/react-hook-form/issues/230#issuecomment-522764265


bluebill1049:
Thanks @dandv

PR #274, opened 2020-05-29
Fixed example code on FormContext Performance

keiya01:
I fixed example code on FormContext Performance of Advanced Usage.

I think `useMemo()` should be used to memorize expensive  value but not a component. 
Therefore I fixed to use `memo()`.

codesandbox: https://codesandbox.io/s/react-hook-form-useform-template-gtx2c?file=/src/index.js

### DIFF
--- a/src/components/codeExamples/formContextPerformance.ts
+++ b/src/components/codeExamples/formContextPerformance.ts
@@ -1,5 +1,23 @@
-export default `import React from "react";
+export default `import React, { memo } from "react";
 import { useForm, FormContext, useFormContext } from "react-hook-form";
+
+// we can use React.memo to prevent re-render except dirty state changed
+const NestedInput = memo(
+  ({ register, formState: { dirty } }) => (
+    <div>
+      <input name="test" ref={register} />
+      {dirty && <p>This field is dirty</p>}
+    </div>
+  ),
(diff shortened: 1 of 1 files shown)


bluebill1049:
agree. your solution looks better.

src/components/codeExamples/formContextPerformance.ts
```diff
@@ -1,37 +1,36 @@
...
   const methods = useForm();
   const onSubmit = data => console.log(data);
 
   return (
     <FormContext {...methods}>
-      // pass all methods into the context
+      {/* pass all methods into the context */}
       <form onSubmit={methods.handleSubmit(onSubmit)}>
-        <NestedInput />
+        <NestedInputContainer />
```
bluebill1049, line 30: make sure you pass down the props :)

keiya01: Do you mean pass `method` variable to NestedInputContainer ?

bluebill1049: wait my bad, let me read the code again.

bluebill1049: never mind, what you got is correct. thanks!

PR #277, opened 2020-05-30
Translated  advanced useage page into jp

keiya01:
I translated Advanced Usage page into jp.


bluebill1049:
Thank you very much!

PR #279, opened 2020-05-31
Fixed data path on ApiPage 

keiya01:
Sorry, the file path were incorrect on ApiPage.


bluebill1049:
aha all good. i was too lazy to check. my bad :(

PR #280, opened 2020-05-31
Fixed bugs on ApiPageV5

keiya01:
I fixed bugs that data is not loaded.


bluebill1049:
thank you so much for fixing those @keiya01 🙏 I really appreciated that!

PR #282, opened 2020-06-01
Split the bundle on each page

keiya01:
I split the bundle on AdvancedUsagePage and GetStartedPage.

And I deleted the files that had become unused due to splitting the bundle.


kotarella1110:
awesome! thanks 💯

PR #284, opened 2020-06-01
fix error message type and update for v6

kotarella1110:
related https://github.com/react-hook-form/react-hook-form/issues/1713


kotarella1110:
@bluebill1049 umm... why deployment has failed?

bluebill1049:
hmmm no idea, i wll get it in and fix it. thanks @kotarella1110

PR #291, opened 2020-06-07
Feature/ts codesandbox

pmaier983:
Fix some of the v6 documentation. Also use resolver instead of validationResolver in v6 codesandbox, in preparation for v6 launch.


bluebill1049:
Thank you so much all for those contributions!

PR #305, opened 2020-06-21
Support i18n in Subscribe Input Change section in Home

keiya01:
I supported i18n in SubScribe Input Change section.
And I translated to jp.


bluebill1049:
awesome! thank you! i completely miss that part.

PR #306, opened 2020-06-21
Translated to jp in FAQs page

keiya01:
- I fixed typo
- I translated to jp


bluebill1049:
awesome! thank you @keiya01 🙏

PR #308, opened 2020-06-23
fix examples with email regex to support all TLD's

sandcastle:
Currently, the examples that have email regex force the TLD (top-level domains) to be in the length range of 2-4 characters. This tripped me up for a little while when testing the sample code as we use "*@xyz.digital" for our emails and caused confusion as to if there was a bug with react-hook-form or the regex.

Hope it saves someone else some time in the future.


bluebill1049:
thanks a lot :)

PR #312, opened 2020-06-25
Feature/api doc work

pmaier983:
- Uncouples generateCode from CodeArea
- Creates a "Types" button (per spectrum discussion)
- re-adds Types back into error options.
- add key prop to .map in VideoList (was Causing Error)

I used a-lot of _.isEqual... this can be improved but at the cost of simplicity and readability and I didn't think the "performance" was worth it... at leas until we have ~10,000 CodeAreas. But am willing to change if prompted!


bluebill1049:
looks great! I think I need to update the design with `copy` button, cause they look too similar

PR #313, opened 2020-06-25
remove lodash from codeArea & enhance logic

pmaier983:
- removes lodash/fp (and lodash entirely) string === string
- add logic to handle just (ss & types) or just (types and js)


bluebill1049:
beautiful! lost some weight!

PR #318, opened 2020-06-27
Expand/improve setError Docs

pmaier983:
- Use best practice ```type``` variable for the first TS example 
- Fix some english issues
- Simplify setError Code Examples
- Create setError TS Code Examples
- Build a large inclusive CodeSandbox for both Ts and JS
	- Both CodeSandbox's have examples for three setError use cases


bluebill1049:
beautiful! thanks 🙉

PR #319, opened 2020-06-28
Fix CodeArea component bugs

keiya01:
I fixed bugs that code examples are not displayed if you access to `register` description section in the API page and you clicked radio button.


bluebill1049:
nice!

PR #323, opened 2020-06-28
Enhance clearError docs + fix bug

pmaier983:
- add TS and JS code example and sandbox to clearError Docs
- fix CodeArea bug that would show CodeSandbox link even when there was none.


bluebill1049:
nice! thank you 🙉

PR #327, opened 2020-06-29
Enhance setValue Docs

pmaier983:
A couple smaller things in this PR.
- remove dead space from format script
- visually de-clutter clearError docs
- add TS, JS and Types + CodeSandbox for setValue


bluebill1049:
awesome, thanks @pmaier983

PR #330, opened 2020-07-01
fix: Fix usage of setValue in V6 migration guide.

paambaati:
I'm assuming you meant to say `setValue` and not `setError`.

Screenshot 2020-07-01 at 3 20 23 PM


bluebill1049:
you are right! thanks

PR #331, opened 2020-07-01
Fixed code example in accessibility section of advanced page

keiya01:
I fixed code example in accessibility section.
Because `aria-description` is not read dynamically. And if you use `role="alert"` at initial rendering, screen reader unintended to read contents.

About `role="alert"`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute


bluebill1049:
This is great! you are right. <3

PR #333, opened 2020-07-01
Use formik hooks example as real apples to applies comparison.

ian:
Seems like it would be best to use formik hooks example to better show comparison between libraries.  

Here's what the overall example looks like: 

```
import React from "react"
import { useFormik } from "formik"
import * as Yup from 'yup';

const validationSchema = Yup.object().shape({
  email: Yup.string().required("Required").email("invalid email address"),
  username: Yup.string().notOneOf(["admin"], "Nice try!")
});

const Example = () => {
  const onSubmit = values => console.log(values);
  const formik = useFormik({
    initialValues: {
      email: '',
      username: '',
    },
    onSubmit,
    validationSchema,
  });

  return (
    <form onSubmit={formik.handleSubmit}>
      <input
        id="email"
        name="email"
        value={formik.values.email}
        onChange={formik.handleChange}
      />
      {formik.errors.email}

      <input
        id="username"
        name="username"
        value={formik.values.username}
        onChange={formik.handleChange}
      />
      {formik.errors.username}

      <button type="submit">Submit</button>
    </form>
  )
};
```

### DIFF
--- a/src/components/V5/codeExamples/formikCode.ts
+++ b/src/components/V5/codeExamples/formikCode.ts
@@ -1,51 +1,43 @@
-export default `import React from "react";
-import { Formik, Form, Field } from "formik";
+export default `import React from "react"
+import { useFormik } from "formik"
+import * as Yup from 'yup';
 
-function validateEmail(value) {
-  let error;
-  
-  if (!value) {
-    error = "Required";
-  } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,4}$/i.test(value)) {
-    error = "Invalid email address";
-  }
-  
-  return error;
-}
-
-function validateUsername(value) {
-  let error;
-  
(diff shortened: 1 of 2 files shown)


bluebill1049:
I do appreciate your putting this together and your love toward Formik (respect that), however this website it's made to highlight RHF. we probably could update the redux-form version as well, but that's not the point to track other libraries otherwise the website would have called xx vs xx vs xx vs. plus what we got are copied from official docs. Also, you should take a look at the comparison closely, it's about to compare build-in validation, not apply a third party lib.

ian:
I just figured if you're going to compare, it should be an _apples to apples_ comparison.

pmaier983:
Hello @ian. You bring up a good point. I'll talk with bill about updating the examples! This is something that needs to be done careful to highlight react-hook-forms strong points. For now I will archiving this PR until it is needed again! Thank you!

PR #335, opened 2020-07-01
update controlled component render code docs

sumeesh879:


bluebill1049:
oooops! thans a lot for that.

PR #336, opened 2020-07-02
Enhance trigger docs

pmaier983:
There will be a fix for [Controller](https://github.com/react-hook-form/documentation/issues/324#issuecomment-652691227) coming up next. This is leftovers from last night.
- add V6 JS and TS examples and CodeSandboxes


bluebill1049:
Thanks! @pmaier983

PR #344, opened 2020-07-02
Transrate controller onFocus docs to jp.

daisuke-awaji:


bluebill1049:
Thank you! we missed it.

PR #346, opened 2020-07-03
fix: css declaration

cpprookie:
image

fix css module type declaration error based on https://stackoverflow.com/questions/41336858/how-to-import-css-modules-with-typescript-react-and-webpack/51536973#51536973


pmaier983:
This is awesome!

PR #348, opened 2020-07-05
Fix/eslint update

pmaier983:
This Commit enables the a11y eslint rules.
- Fixes a ton of general eslint errors (509 errors -> 43)
- Fixes all the major a11y errors


src/components/SideMenu.tsx
```diff
@@ -145,7 +145,10 @@ function SideMenu({
                 delay={index * 0.01}
                 render={({ style }) => (
                   <li
+                    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-to-interactive-role
+                    role="link"
                     onClick={() => goToSection(enLink, index)}
+                    onKeyDown={() => goToSection(enLink, index)}
```
keiya01, line 151: I think this action should executed only when `Enter` is clicked.

src/components/V5/ApiFormStateV5.tsx
```diff
@@ -133,8 +133,11 @@ export default React.memo(
                     <b className={typographyStyles.note}>Note:</b>{" "}
                     <code>isValid</code> is affected by{" "}
                     <code
+                      role="link"
+                      tabIndex={0}
                       className={buttonStyles.codeAsLink}
                       onClick={() => goToSection("useForm", sectionsRef)}
+                      onKeyDown={() => goToSection("useForm", sectionsRef)}
```
keiya01, line 140: This code is The same reason too.

bluebill1049:
Thanks @keiya01 for the review.

src/components/ApiPage.tsx
```diff
@@ -275,7 +275,7 @@ function ApiPage({ formData, defaultLang, api }: Props) {
             {links.map((option) => {
               const title =
                 typeof option === "function" ? option().title : option.title
-
+              console.log("the title:")
```
bluebill1049, line 278: remove?

pmaier983: good catch!

bluebill1049:
Thanks @pmaier983

src/components/V5/ApiPageV5.tsx
```diff
@@ -338,90 +338,129 @@ function ApiPage({ formData, defaultLang, api }: Props) {
...
+                tabIndex={0}
               >
                 register
               </code>
               ,{" "}
               <code
                 className={buttonStyles.codeAsLink}
                 onClick={() => goToSection("unregister")}
+                role="link"
+                onKeyPress={() => goToSection("unregister")}
```
keiya01, line 352: I think `onKeyPress` events should be performed with only `Enter` too. And I think these code link tag should be wrapped `a` tag. Because it can perform `onKeyPress` and `focus` with default.
Thank you 🙏

pmaier983: Thank you for the insight! I clearly need to do some more research into a11y instead of just making the linter happy

pmaier983: I updated the Code snippets to activate on enter click. This along with the tabIndex will allow someone to tab through the options and enter on the one they like (try it out, it works!). I did not wrap in an <a> tag. We handle intradoc linking in an interesting unique way that does not make use of `<a href="#tag">`. Instead we use a goToSection function that also allows us to comply with i18n.

bluebill1049: oh wow! nice! 🚀

keiya01: thank you @pmaier983 😄

PR #350, opened 2020-07-06
Improve performance in api page

keiya01:
I fixed css and performance with Register Options feature in API page.

- I added `cursor: pointer;`
- I replaced from `Prism.highlightAll` to `Prism.highlightAllUnder` because it is too late when I clicked Register Options.


bluebill1049:
wow nice instant now! 🤟

PR #351, opened 2020-07-06
improved render-blocking

keiya01:
Related to #350, I fixed render-blocking in `Layout` components. To fix this, I move `Prism.highlightAll` to `CodeArea` component and I replaced to `Prism.highlightAllUnder`. This problem occur when it toggle light mode and dark mode or it navigate another page.

### DIFF
--- a/src/components/CodeArea.tsx
+++ b/src/components/CodeArea.tsx
@@ -2,6 +2,7 @@ import * as React from "react"
 import copyClipBoard from "./utils/copyClipBoard"
 import { useStateMachine } from "little-state-machine"
 import generic from "../data/generic"
+import Prism from "prismjs"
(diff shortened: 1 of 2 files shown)


src/components/CodeArea.tsx
```diff
@@ -84,6 +86,15 @@ export default function CodeArea({
   const { currentLanguage } =
     language && language.currentLanguage ? language : { currentLanguage: "en" }
 
+  React.useEffect(() => {
+    const highlight = async () => {
+      if (codeAreaRef.current) {
+        Prism.highlightAllUnder(codeAreaRef.current)
```
bluebill1049, line 92: nice! that's the bit i didn't know ❤️ Awesoem @keiya01

PR #353, opened 2020-07-07
fix: language select state

cpprookie:
sync select state with location.

Steps to reproduce bug.
- click  https://react-hook-form.com/api
- click useForm v5 and select chinese
- click v6 then navigate to english api page but select value not update because of defaultValue props.


bluebill1049:
👍 nice one.

Issue #352, opened 2020-07-06
ErrorMessage multiple errors


MauriceOppenberger:
The Documentation of displaying multiple errors does not seem to work properly when using TypeScript
@ [api#ErrorMessage](https://react-hook-form.com/api#ErrorMessage) 
```
<ErrorMessage errors={errors} name="multipleErrorInput">
        {({ messages }) =>
          messages &&
          Object.entries(messages).map(([type, message]) => (
            <p key={type}>{message}</p>
          ))
        }
      </ErrorMessage>
```

Error: 
Property 'children' does not exist on type 'IntrinsicAttributes & { as?: undefined; errors?: DeepMap<inputFields, FieldError>; name: string; message?: string; render?: (data: { ...; }) => ReactNode; } & Pick<...>

solution: 
following npm packages documentation 
@ [npm/hookform](https://www.npmjs.com/package/@hookform/error-message)
```
<ErrorMessage
        errors={errors}
        name="multipleErrorInput"
        render={({ messages }) =>
          messages &&
          Object.entries(messages).map(([type, message]) => (
            <p key={type}>{message}</p>
          ))
        }
      />
```

pmaier983:
I'll investigate whats going on here, and later today I'll see if I can update the example with some TS code and a CodeSandbox. Thank you for bringing this up!

pmaier983:
Take a look at the CodeSandbox's created in https://github.com/react-hook-form/documentation/pull/354. I believe they provide a working example of TS and ErrorMessage!

pmaier983:
Closing this issue, but feel free to re-open if you have a specific CodeSandbox detailing the error! Thank You!

PR #354, opened 2020-07-07
Fix/error messages ts issue

pmaier983:
This PR updated the examples for ErrorMessage to adress issue #352. 
- Adds TS and JS codeExamples and Code for both multi and single error.


bluebill1049:
Thanks @pmaier983 👍

PR #357, opened 2020-07-08
Add useWatch Examples

pmaier983:
Added TS, and JS CodeSandbox and a TS example. Slow but steady 😄!


bluebill1049:
nice! 🍰 thanks @pmaier983

PR #358, opened 2020-07-09
feat(smart-form): use React.Children.map()

pcooney10:
feat(smart-form): use React.Children.map()
- works with a single child + an array of children
- docs: https://reactjs.org/docs/react-api.html#reactchildrenmap


bluebill1049:
nice one!

pcooney10:
Wow that was fast ! 🏎️

PR #359, opened 2020-07-09
Update copy and dynamically create refs on GetStartedPage

msutkowski:
This PR: 

- Updates various copy throughout the docs for readability
- Dynamically creates refs on sections for `scrollTo` purposes

Original comment: 
Great work on the library! I was reading through the docs and found a handful of things that could be adjusted for readability. In doing so, I ran into an issue with the refs on the `GetStartedPage` and thought to create the dynamically as a title change would be problematic in the future. That brings up another problem of permalinks regarding the hash routes on the get-started doc. Figured I'd open this and see if you all had any thoughts about it.


bluebill1049:
This is awesome, sorry i have been pretty lazy with this for a while, is it possible to build this so it's reusable across the site?

msutkowski:
@bluebill1049 This is still a WIP, but I just updated the comment about the issue regarding permalinks. I need to do another once over in general, but this is a decent start. Let me know if you have any input so far and I'll wrap this up tomorrow.

bluebill1049:
> @bluebill1049 This is still a WIP, but I just updated the comment about the issue regarding permalinks. I need to do another once over in general, but this is a decent start. Let me know if you have any input so far and I'll wrap this up tomorrow. I could map all previous permalinks that went to the same section (but renamed), but it could get ugly. I'll have to think more on it.

sounds good to me, no rush on this, take your time. You saving us heaps of time in the near future. ❤️

.gitignore
```diff
@@ -27,3 +27,4 @@ yarn-error.log*
 .idea
 .cache/
 package-lock.json
+.vscode/settings.json
```
pcooney10, line 30:
```suggestion
.vscode
```

src/data/en/faq.tsx
```diff
@@ -591,14 +595,15 @@ export default {
       description: (
         <>
           <p>
-            It's important to understand React Hook Form embrace native form
-            behavior by store input state inside each input (except custom
+            It's important to understand React Hook Form embraces native form
+            behavior by storing input state inside each input (except custom
```
msutkowski, line 599: Could use a once over on the exception language

msutkowski:
I'm not sure what happened with the merge here - there was only one small block that got pulled in. Let me know your thoughts, and thanks!

bluebill1049:
> I'm not sure what happened with the merge here - there was only one small block that got pulled in. Let me know your thoughts, and thanks!

I will sort out the conflicts. thanks for the awesome update.

bluebill1049:
actually, you want to try to merge master into your branch and resolve the conflict there.

msutkowski:
I thought I did, but who knows. I'll resolve it in the morning 🎉

msutkowski:
> LGTM! Awesome contribution! 🎆 🎆

@pmaier983 Thanks! The only thing I'm still debating on or concerned with is the scrollTo behavior on hash links. I don't really have a good solution for backward compatibility on links (from a permalink perspective) if that's a concern for the RHF team at this time. In general, it seems strange to tie the hashes like `get-started#namefromtitleproperty`, because as the titles are changed in the future it'd potentially "break" backlinks. It might make more sense to tie them to the key (such as `controlledInput` in `getStarted`)  instead of the `getStarted.controlledInput.title` or to set a hash value property as well as a `title` that isn't meant to be permanent. I suppose the worst-case scenario with the implementation this offers currently is that a user would still be on the correct page, and they'd just have to use the menu. 

I think that's all that anyone needs to be aware of at this point. If the team cares about 100% backlinkability _(or whatever a good term for that is)_, I can also add in a resolver for unfound hash paths and try to match them to previously known ones... but that might be overkill. I'm open to any suggestions regarding this :)

bluebill1049:
> > LGTM! Awesome contribution! 🎆 🎆
> 
> @pmaier983 Thanks! The only thing I'm still debating on or concerned with is the scrollTo behavior on hash links. I don't really have a good solution for backward compatibility on links (from a permalink perspective) if that's a concern for the RHF team at this time. In general, it seems strange to tie the hashes like `get-started#namefromtitleproperty`, because as the titles are changed in the future it'd potentially "break" backlinks. It might make more sense to tie them to the key (such as `controlledInput` in `getStarted`) instead of the `getStarted.controlledInput.title` or to set a hash value property as well as a `title` that isn't meant to be permanent. I suppose the worst-case scenario with the implementation this offers currently is that a user would still be on the correct page, and they'd just have to use the menu.
> 
> I think that's all that anyone needs to be aware of at this point. If the team cares about 100% backlinkability _(or whatever a good term for that is)_, I can also add in a resolver for unfound hash paths and try to match them to previously known ones... but that might be overkill. I'm open to any suggestions regarding this :)

I couldn't agree more with above, I think moving forward that the solution we want anyhow. Having a resolver for unfound hash can be step as well, it's a great idea and we can introduce some good UX to fulfill the gap when content is not found.

PR #361, opened 2020-07-10
Cleanup before a heavy lift

pmaier983:
- Cleanup some dead filed
- Updated package dependencies to the latest and greatest (this was done to avoid as many bugs as possible going forwards)
- Bump react-hook-form version 6.0.2 yay!
- Bump node version 12 -> 14 (12 will eventually be unsupported)(I can change it back if anyone wants, just a pet peeve of mine #nvm)
- Delete Global.d.ts (it wasn't doing what it was supposed to)

P.S. I'm hoping this weekend I'll have the time to work on a larger refactor of the api page code. This was some cleanup in preparation. More info to come.

### DIFF
--- a/Global.d.ts
+++ b/Global.d.ts
@@ -1,1 +1,0 @@
-declare module "*.module.css"
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 ## Install
 
-    npm i && npm start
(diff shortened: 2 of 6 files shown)


bluebill1049:
looks like the build failed :(

PR #365, opened 2020-07-11
BigChanges

pmaier983:
This is less of a PR and more of a suggested restructuring (all input is welcome)
What this PR does:
- Increases colocation of codeExamples and translations to use case (this improves dev experience)
- Translates some stuff (using [deepl](https://www.deepl.com/home))
- Creates a new method for handling translations (allows for partial translations as it default to english)


src/components/utils/languages.ts
```diff
@@ -0,0 +1,46 @@
+import { has, get } from "lodash"
+
+export const SUPPORTED_LANGUAGES = {
```
cpprookie, line 3: `enum` and `Map` may behave better than object literal for type annotation.

pmaier983: I fully agree! Will change 😄

pmaier983:
If everyone is ok with what this branch proposes, I'll start converting the full API page, and then move onto other pages over time. If you think we should be doing more of X or Y, now is a good time to bring it up so that it can happen over the next couple months as I comb through and convert!

keiya01:
That's great! However I think this changes have possible to increase bundle size. This is tradeoff which we take either dev experience or performance ...

bluebill1049:
> That's great! However I think this changes have possible to increase bundle size. This is tradeoff which we take either dev experience or performance ...

good point! I would vote towards user experience (which is perf). Also, it's probably worth measuring the diff as well.

bluebill1049:
**Example:**

one of the mistakes I did earlier was using styled component, it's a great DX (i still love it), but it's a price for each user to pay, which is runtime cost for the app, so I converted the whole app to CSS, it's painful but I rather observe the pain myself in exchange for a faster web for users.

pmaier983:
I definitely agree that we should value user experience > developer experience, and its awesome that this is something we care alot about as a Team! But I don't think it is a zero sum game. User's have a threshold for acceptable experience, I believe google uses 2.5s and below as a "good" amount of time for a websites largest contentful paint (We are currently well below this and I want it to stay that way!). 

<details><summary>Lighthouse Score</summary>
Screen Shot 2020-07-12 at 10 15 08 AM
</details>

The proposed changes will impact performance for sure, but the impact will likely be milliseconds at most. The two main speed effecting changes in this PR are: more object declarations, more function calls. Both things javascript and modern browsers can do very well. 

What I'm saying is that I am willing to trade off a marginal amount of speed (so long as it doesn't negatively effect users) for developer experience. In addition this PR makes it possible for half translations to easily exist, where a translation will default to english if no one has translated it yet. This significantly improves i18n, and makes any millisecond losses worth it.

I am 100% open to suggestions on how best to improve performance, and I definitely agree that we should keep a close watch on performance in every PR (I wonder if there is a CI tool that can help with this?).

bluebill1049:
> I definitely agree that we should value user experience > developer experience, and its awesome that this is something we care alot about as a Team! But I don't think it is a zero sum game. User's have a threshold for acceptable experience, I believe google uses 2.5s and below as a "good" amount of time for a websites largest contentful paint (We are currently well below this and I want it to stay that way!).
> 
> Lighthouse Score
> The proposed changes will impact performance for sure, but the impact will likely be milliseconds at most. The two main speed effecting changes in this PR are: more object declarations, more function calls. Both things javascript and modern browsers can do very well.
> 
> What I'm saying is that I am willing to trade off a marginal amount of speed (so long as it doesn't negatively effect users) for developer experience. In addition this PR makes it possible for half translations to easily exist, where a translation will default to english if no one has translated it yet. This significantly improves i18n, and makes any millisecond losses worth it.
> 
> I am 100% open to suggestions on how best to improve performance, and I definitely agree that we should keep a close watch on performance in every PR (I wonder if there is a CI tool that can help with this?).

sounds good to me, especially the measurement, let's moving forward and taking the required adjustment if we need it after.

keiya01:
> sounds good to me, especially the measurement, let's moving forward and taking the required adjustment if we need it after.

we must find between DX and UX.

> I am 100% open to suggestions on how best to improve performance, and I definitely agree that we should keep a close watch on performance in every PR (I wonder if there is a CI tool that can help with this?).

I think we can use [lighthouse-ci](https://github.com/GoogleChrome/lighthouse-ci).
We should probably improve while checking performance by using CI.

pmaier983:
Ill take a look into adding lighthouse-ci this week! Great idea!!

bluebill1049:
Thanks @pmaier983 and @keiya01 ❤️

bluebill1049:
Aim for the best doc for our developer friends! 🎉

pmaier983:
We now have some Awesome lighthouse-ci based checks: 

https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1594852470722-96645.report.html
https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1594852471021-972.report.html
https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1594852471309-9246.report.html
https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1594852471761-70344.report.html
https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1594852472126-65836.report.html

One for every html page.
ToDo in the future: Make a server to record the tests and track progress

P.S currently if any of the pages has a First contentful paint > 2.5s the merge is blocked (more assertions like this will be added following a discussion)

bluebill1049:
nice!

pmaier983:
Hey Bill, What do you think this PR needs to push it over the line? Should this PR convert the entire page in this manner? If so, I'll start committing, but it will turn into a rather large PR (Not that it's an issue)

bluebill1049:
let's do it progressively if you think it's ready let's review it. 👍

pmaier983:
I hesitantly proclaim this PR to be ready... please find all the issues!!

bluebill1049:
looks fine to me 👍

src/components/api/useWatch/UseWatch.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <code className={typographyStyles.codeHeading}>
+      <h2>
+        useWatch:{" "}
+        <span
+          className={typographyStyles.typeText}
+        >{`({ control?: any, name?: string, defaultValue?: any }) => object`}</span>
+      </h2>
+    </code>
+
+    {chooseTranslation(translation, currentLanguage, "description")}
```
keiya01, line 29: Why do you not access like `translation[currentLanguage].description` ? If we can access like this, we can take benefit by input completion.

src/components/utils/languages.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const chooseTranslation = (
+  translation,
+  language,
+  path,
+  defaultLanguage = "en"
+) =>
+  has(translation, `${language}.${path}`)
+    ? get(translation, `${language}.${path}`)
+    : get(translation, `${defaultLanguage}.${path}`)
```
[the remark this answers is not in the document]
pmaier983: I chose a default language because we often don't have full translations. Instead of copy-pasting (and making it more difficult to make changes accurately) I think it would be best to default to english, and show all languages side by side. This will also hopefully make it easier to translate!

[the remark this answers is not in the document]
keiya01: I understand. That's good!
However I think we make be able to useful if we can use type inference. Is this difficult ?

src/components/api/useWatch/UseWatch.tsx
keiya01: Thank you.

src/components/utils/languages.ts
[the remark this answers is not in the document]
pmaier983: I am no TS master, but i'll take a shot at it :D (Good Point!)

pmaier983:
@keiya01. Sorry for the long wait! So I tried my hand at giving the handleTranslation function a solid type inference... 

```javascript
export const chooseTranslation = (
  translation: { [Key in SUPPORTED_LANGUAGES]?: React.ReactNode | string },
  language: SUPPORTED_LANGUAGES,
  path: string,
  defaultLanguage = "en"
): React.ReactNode | string =>
  has(translation, `${language}.${path}`)
    ? get(translation, `${language}.${path}`)
    : get(translation, `${defaultLanguage}.${path}`)
```

I am unsure in my use of `React.ReactNode | string`. What I mean is the return is either a node `<div>hello</div>` or a string. Is React.ReactNode best practice?

pmaier983:
Yea i have been looking into gatsby i18n. Hoping to have a POC soonish. (feel free to try your own hand tho!)

keiya01:
@pmaier983 Thank you very much for fixing types. This is awesome :)

pmaier983:
I'm going to close this PR until I'm certain it doesn't conflict with `gatsby-theme-i18n`. Preliminary exploration indicates this PR and the new i18n library conflict ☹️. I'll keep everyone up-to-date tho!

PR #372, opened 2020-07-14
Fix advanced-usage path on ApiPage

valmassoi:
Clicking the link when the current location is something like `/api/#formState` would result in a 404 page `/api/advanced-usage`


bluebill1049:
ty

Issue #375, opened 2020-07-15
Wrong menu item underlined


michalbundyra:
Go to: https://react-hook-form.com/api/#useWatch
`useFieldArray` menu item is underlined instead of `useWatch`

<img width="349" alt="image" src="https://user-images.githubusercontent.com/7423207/87554036-e87a5b80-c6b3-11ea-939f-dc3abc118bd8.png">

Go to: https://react-hook-form.com/api#useFieldArray
`useWatch` menu item is underlined instead of `useFieldArray`

<img width="341" alt="image" src="https://user-images.githubusercontent.com/7423207/87554118-05169380-c6b4-11ea-8c16-9abfa88ed935.png">

bluebill1049:
thanks @michalbundyra

pmaier983:
you taking care of this one [Mr.Bill](@bluebill1049)?

bluebill1049:
could you take a look at this one? @pmaier983 it's not trivial

pmaier983:
The following PR fixes the issue: https://github.com/react-hook-form/documentation/pull/376

PR #376, opened 2020-07-16
Simplify useEffect and fix navigation highlighting

pmaier983:
This PR fixes Issue #375.

- The real issue is that the refs were declared in an improper order... ([Lines 166 and 167](https://github.com/react-hook-form/documentation/compare/master...pmaier983:fix/navigation-highlighting?expand=1#diff-8e2137ec74d914a55499d50c2858f9bcL166)) but to find this I first refactored the entire highlighting useEffect to be more functionally coded. What I'm saying is, this PR could be a 2-liner if the refactoring is not up to snuff.


bluebill1049:
bravo!

bluebill1049:
tested and works!

PR #380, opened 2020-07-18
Change `criteriaMode` defalutValue from `firstErrorDetected` to `firstError`

matamatanot:
https://github.com/react-hook-form/react-hook-form/blob/d93e4427cf6c60ea76ac1e3a30e24e8e3ef2f399/src/types/form.ts#L101


bluebill1049:
thank you @matamatanot

Issue #381, opened 2020-07-18
Translated Side Menu fragment link is not working


matamatanot:
Working
https://react-hook-form.com/get-started#ReactWebVideoTutorial

Not Working
https://react-hook-form.com/jp/get-started#ReactWebVideoTutorial
https://react-hook-form.com/pt/get-started#ReactWebVideoTutorial
etc...

-----

I read a little bit of code. It seems to be caused by `esLinks`.
https://github.com/react-hook-form/documentation/blob/0db20b179ba423b7659a93b2779687632dc58b2c/src/components/GetStartedPage.tsx#L37

I modified part of it as follows and it worked. This is an example, not an abstraction.

https://github.com/react-hook-form/documentation/blob/0db20b179ba423b7659a93b2779687632dc58b2c/src/components/SideMenu.tsx#L73

```
let enLink =
              typeof enLinks[index] === "function"
                ? enLinks[index]().title
                : enLinks[index].title

enLink =
              enLink === "Quick start" ? "クイックスタート" : enLink
```

pmaier983:
Great find matamatanot! Would you like to open up a PR to solve this bug? If not ill fix it in the near future.

pmaier983:
I'll be taking a look into this in the next few days FYI

bluebill1049:
Thanks @pmaier983 👍

matamatanot:
Thanks!

pmaier983:
This Issue was closed when PR #385 was merged.

PR #385, opened 2020-07-23
Fix Side Menu navigation to resolver tab

pmaier983:
This PR fixes Issue #381. 
- Some minor chinese translation (tribute to GF 😉)
- Fix the title... don't love that the title is used as the key for goToSection... may change this in the future.


bluebill1049:
> This PR fixes Issue #381.
> 
> * Some minor chinese translation (tribute to GF 😉)
> * Fix the title... don't love that the title is used as the key for goToSection... may change this in the future.

yes, it's a tech debt for me to resolve, but just been lazy...

PR #393, opened 2020-08-02
fix: use getstartedEn title

cpprookie:
bug fix.
getStarted page menu cannot navigate to right place except english titles currently because of keys  in `sectionRef` uses other language chars while `goToSection` uses an enlingsh one.

### DIFF
--- a/src/components/GetStartedPage.tsx
+++ b/src/components/GetStartedPage.tsx
@@ -145,7 +145,7 @@ const Faq = ({ location, defaultLang, getStarted }: Props) => {
             getStarted={getStarted}
             quickStartRef={(ref) => {
               sectionsRef.current[
-                getRefNameFromTitle(getStarted.install.title)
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks @cpprookie

cpprookie:
> Thanks @cpprookie
My pleasure, @bluebill1049 . By the way, how is i18n refactor going on? Shall i wait for you guy's refactor to translate chinese version or do it in parallel?

pmaier983:
Def wait for bill's answer. But I support in parallel. Getting translations out is more important than a tech change (in this case imo).

bluebill1049:
I am happy to merge what @pmaier983 got :) let's translate in parallel and try to use gatsby official i18n plugin after :)

cpprookie:
Ok, i will work on chinese version step by step, @bluebill1049 @pmaier983. I thought i18n refactor would break up current files structure as a huge change before so just waited for it. LOL

PR #397, opened 2020-08-05
Fix code example in Controller docs

msutkowski:
Fixes a typo on the code example. `props` isn't available as everything is destructured.


bluebill1049:
Thank you @msutkowski !

PR #398, opened 2020-08-06
Update formWizard.ts

cpprookie:
fix: `submit` should fire in `Result` component according to route config.


bluebill1049:
Thank you @cpprookie

PR #400, opened 2020-08-06
chore: enhance advanced cn

cpprookie:
Terms translation:
- advanced -> 进阶
- form field -> 表单项  based on antd usage.

Better chinese expression.
Update legacy parts to sync with en version.


bluebill1049:
thank you @cpprookie !

PR #405, opened 2020-08-11
include ie11 info

bluebill1049:
related to : https://github.com/react-hook-form/react-hook-form/pull/2551/files

### DIFF
--- a/src/data/en/faq.tsx
+++ b/src/data/en/faq.tsx
@@ -147,12 +147,13 @@ export default {
       title: "Browser Support?",
       description: (
         <>
-          <p>React Hook Form supports all major browsers.</p>
-
           <p>
-            For legacy IE11 support, you can import react-hook-form IE 11
-            version.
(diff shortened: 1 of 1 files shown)


pmaier983:
The lighthouse test is a bit flakey. I'll increase its timeout.

bluebill1049:
would be nice if the lighthouse can produce a report, like some of the action that we using in the core lib.

pmaier983:
Actually Lighthouse seems to be failing at the build step... Mb it was a fluke. If this issue: (`Error: Cannot find module 'gatsby-cli/lib/reporter'`) shows up again I'll have to fix the action.

pmaier983:
It actually does create a report! Its just hidden in the github action log. 
Screen Shot 2020-08-11 at 9 48 34 PM
The links expire quickly though, but you can provide a server to store them. This is a plan of mine down the road!

PR #409, opened 2020-08-20
Updated onTouched method

alexw23:
It seems the documentation seems to imply Validation will only trigger until an input has been touched "Validation will trigger until input is touched". This should be "won't" trigger until and input is touched.


bluebill1049:
Thank you! much better.

PR #411, opened 2020-08-23
Update Strictly Typed code example with correct implementation of using checkbox with TypeController

FrimJo:
According to discussion in this issue [Checkbox when using `useTypedController` not working as expected](https://github.com/react-hook-form/react-hook-form/issues/2678#issue-683874150) on the `react-hook-form` repository. This PR updates the code example on how to use Strictly Typed with implementation of checkbox to be more inline with example on how to use Controlled checkbox.


bluebill1049:
Thank you very much.

PR #414, opened 2020-08-28
Add id to all "goToSection" tags

pmaier983:
This PR adds id's to as many anchored links as I could find. This will allow docSearch to better map to them, because currently it pastes `#gatsbyfocuswrapper` instead of anything useful.


bluebill1049 (approved):

bluebill1049:
Thank you.

PR #416, opened 2020-08-31
Feature/search bar

bluebill1049:
- [x] style the search section.


src/components/Nav.tsx
```diff
@@ -26,98 +28,105 @@ export default function Nav({ defaultLang }: { defaultLang: string }) {
...
+  React.useEffect(() => {
+    if (setting.isFocusOnSearch && window.innerWidth < 767) {
+      setLang(false)
+    } else if (!setting.isFocusOnSearch) {
+      setLang(true)
+    }
+
+    if (
+      setting.isFocusOnSearch &&
+      window.innerWidth < 767 &&
```
pmaier983, line 40: i think in the future we should have a breakpoint object somewhere. Each of the values mapping to a normal device (iphone, normal monitor and the such) (@TechDebt)
```javascript
const breakpoints = {
	xsm: 500,
	sm: 756,
	m: 1000,
	l: 1500,
	xl: 2000,
}
```

src/components/layout.tsx
```diff
@@ -1,6 +1,5 @@
 import * as React from "react"
 import { Animate } from "react-simple-animate"
-import Prism from "prismjs"
```
pmaier983, line 3: I need to configure our linter so that it catches these!

src/components/Search.tsx
```diff
@@ -1,17 +1,65 @@
...
+import * as React from "react"
+import { useStateMachine } from "little-state-machine"
+import { updateSetting } from "../actions/settingActions"
 import searchStyles from "./Search.module.css"
 
 const Search = () => {
+  const timer = React.useRef<any>({})
+  const { action, state } = useStateMachine(updateSetting)
+
+  React.useEffect(() => {
```
pmaier983, line 10: We need to pick gatsby config or this. I don't think both is necessary.

gatsby-config.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
       options: {
         apiKey: "953c771d83fb6ffd55fe58da997f2d9d",
         indexName: "react-hook-form",
-        inputSelector: "#algolia-doc-search",
+        inputSelector: "#fakeSearch",
```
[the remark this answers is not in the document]
bluebill1049: the gatsybe thing doesn't work... it's targeting a fark searh, take a look the search component.

src/components/Search.tsx
bluebill1049: gatsbye one doesn't work when you navigate around the page, try it.

src/components/layout.tsx
bluebill1049: oh i see. will roll it back.

bluebill1049:
> I really love the design here! I think we still need to work on implementation a bit, as on my computer the deploy is a bit janky.
> 
> image
> 
> In the image above you can click on either bubble and it will open up a different search.
> The magnifying glass is not centered on the first bubble, and the second bubble has nothing in it.
> 
> Overall awesome idea for the design Bill! If you need any help ping me 😄 (you must have your hands full with both this and bugs!)

going to fix that, doesn't looks like in local 👍

PR #417, opened 2020-09-02
Use `htmlFor` React element attribute in code example

knowler:
Just a small correction if people are going to be copying this code.

Update: from a quick search, I believe these are the only two.


pmaier983 (approved):
Great change knowler! thank you!

PR #423, opened 2020-09-15
Nit: Fix stray markdown backticks?

mlissner:
Not sure on this one, but I'm guessing at some point somebody translated this from markdown, and missed these backticks when they did so.


bluebill1049:
Thank you!  @mlissner

PR #431, opened 2020-09-27
Replace duplicated `singleValue` with more proper const name

bozheville:


pmaier983 (approved):
Good Catch bozheville!

bluebill1049:
Thanks @bozheville

PR #433, opened 2020-10-01
Fix centering and padding lang select

NextNebula:
Some small fixes for centering and padding for the language select.
- Vertical align selected language.
- Fix caret icon and longer language names overlap.

Current styling:
afbeelding

Fixed styling:
afbeelding


bluebill1049:
thank you again 🙏

PR #434, opened 2020-10-01
Hide edit button properly

NextNebula:
If scrolled to the top of a page the edit and scroll buttons are hidden, but if you hover over the hidden element it still changes the cursor and clicking also works. This PR hides the buttons properly.


bluebill1049:
thank you!

PR #443, opened 2020-10-08
add label to accessible example

apdrsn:
Hi :)
Just wanted to add the label to make the input accessible. Think you forgot as the other a11y examples already have a label.

(Let me know if Master is not the correct target branch)


bluebill1049:
thank you 🙏

PR #446, opened 2020-10-11
clarify the meaning of mode="onTouched"

### LINKED ISSUE #445: API documentation: description of mode="onTouched" is unclear
At the the moment the API documentation on mode="onTouched" of useForm says:

    Validation won't trigger until input is touched (has lost focus).

As a newcomer this was rather unclear for me - especially I couldn't figure out the difference between "onTouched" and "onBlur".

After experimenting a bit I would describe the behavior like this:

    Validation will trigger on the first blur. After that, it will trigger on every change.

(the documentation page is here: <https://react-hook-form.com/api/#useForm> )

marko-knoebl:
fixes #445

### DIFF
--- a/src/data/en/api.tsx
+++ b/src/data/en/api.tsx
@@ -107,7 +107,10 @@ export default {
     ),
     validationOnTouched: (
       <>
-        <p>Validation won't trigger until input is touched (has lost focus).</p>
+        <p>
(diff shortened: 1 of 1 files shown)


bluebill1049:
👍 thanks

PR #450, opened 2020-10-21
fix typescript type in Register fields example

micmro:
`String` needs to be `string` for the _Register fields_ example to compile


pmaier983 (approved):
Good catch micmro! Looks like I'll have to fix lighthouse build. idk what went wrong this time.

PR #455, opened 2020-11-05
add a note to getValues

capaj:
hopefully I worded it correctly


bluebill1049:
That's awesome! thanks

PR #457, opened 2020-11-06
Advanced - Accessibility (A11y): add name attribute to fix the forms

coralieco:
Hi !

When working on form validation with [Advanced - Accessibility (A11y)](https://react-hook-form.com/advanced-usage), I realised the forms for accessibility do not work as it misses a name attribut to the input tag. 

See error in console below:
invalid-form

With my fix, the validation on form works properly:
valid-form


bluebill1049:
oh oops, thanks very much @coralieco

PR #458, opened 2020-11-07
Fix invalid example defaultValue to name

stefanoruth:
To find this goto, https://react-hook-form.com/api#useFieldArray and the example is just below the text: `Important: Because each input can be uncontrolled, id is required with mapped components to help React to identify which items have changed, are added, or are removed.`

### DIFF
--- a/src/data/en/api.tsx
+++ b/src/data/en/api.tsx
@@ -1107,7 +1107,7 @@ React.useEffect(() => {
               rawData={`{fields.map((data, index) =>
   <input
     key={data.id}
-    defaultValue={\`data[\${index}].value\`}
+    name={\`data[\${index}].value\`}
   />;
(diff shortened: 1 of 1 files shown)


src/data/en/api.tsx
```diff
@@ -1107,7 +1107,7 @@ React.useEffect(() => {
               rawData={`{fields.map((data, index) =>
   <input
     key={data.id}
-    defaultValue={\`data[\${index}].value\`}
+    name={\`data[\${index}].value\`}
```
bluebill1049, line 1110:
```suggestion
    defaultValue={\`data[\${index}].value\`}
    name={\`data[\${index}].value\`}
```

bluebill1049:
thanks a lot.

PR #461, opened 2020-11-13
Improve code comparison section

bluebill1049:


src/data/home.tsx
```diff
@@ -273,21 +273,13 @@ export default {
...
-            <b className={typographyStyles.note}>⚠ Note:</b> Both Formik and
-            Redux-Form snippets below were copied from the official
-            documentation.
-          </p>
-        </>
+        <p id="codeComparison">
+          Reducing the amount of code that you have to write is one of the
+          primary goals of React Hook Form. To illustrate that, let's look at
+          this following code example.
+        </p>
```
pmaier983, line 282:
```suggestion
        <p id="codeComparison">
          Reducing the amount of code you need to write, and removing
          unnecessary re-renders are some of the primary goals of React Hook
          Form. Now dive in and explore with the following example:
        </p>
```

pmaier983: I tested this locally and it causes the `example:` to be on its own in the final line, so I suggest the following change as well:

// container.module.css (Lines 24-27)
```
.centerContent p {
  max-width: 720px;
  margin: 0 auto 10px;
}
```

pmaier983 (approved):
I think the CodeSandbox is a great value add! Look awesome!

bluebill1049:
@pmaier983 thanks and feel free to introduce the css update as well :) then we can merge and go live. U should have that CSB access as well

bluebill1049:
thanks a lot, @pmaier983 deploying 🚀  I will close the other PR as well.

PR #462, opened 2020-11-14
Fix typo

upendra-web:
Change a tabname in `reset` api example page from `Submit with rest` => `Submit with reset`


bluebill1049:
ty :)

Issue #468, opened 2020-11-26
Can you store Dark/Light mode settings for the docs in local storage to persist for users?


idrisakmal:
**Is your feature request related to a problem? Please describe.**
I have to set the docs to light mode everytime I load the page. Would be nice to have it persist for each user's settings

**Describe the solution you'd like**
Light/Dark mode settings persist for each user 

**Describe alternatives you've considered**
store the settings in local storage

pmaier983:
this Issue was fixed by #469 😄 Thank you @idrisakmal!

idrisakmal:
thank you @pmaier983 !

amankkg:
![image](https://user-images.githubusercontent.com/3933028/100567871-c1839580-32f3-11eb-9443-68c8a579c4ea.png)
This what I see on every visit. Even if I toggle it to the dark mode on the next visit it's light and links are invisible

bluebill1049:
hmmm looks like it's still working...

bluebill1049:
really strange, that's not even dark or light mode.

bluebill1049:
looks like dark mode is set correctly.

pmaier983:
The issue seems to be relegated to the header. I am able to create it when I go to the official website after clearing all the sites data locally, but I am unable to replicate the issue locally.

bluebill1049:
strange... I will spend some time tonight on this issue. plan b just update in local storage and keep away from the auto detect.

pmaier983:
So I did some sleuthing, It seems on initial load there is some confusion as to what theme to display in. It's almost as if some components cant actually see/use `state?.setting?.lightMode`

bluebill1049:
@amankkg let us know if this is fixed. Thanks @pmaier983's fix!

pmaier983:
I think we should force everyone to default to dark mode (the way it worked before) because as it is now, a new user logging in sees some very odd colors. I'll keep investigating the reason in the background, but this would be a good quick fix.

bluebill1049:
I agree @pmaier983 I did some investigation as well. It's a strange one probably something i did wrong by using Gatsbye. i even debug prod build and all the flag was correct.

pmaier983:
I'm going to be closing this issue now. We are currently defaulting to dark mode as of this commit: https://github.com/react-hook-form/documentation/commit/52590ce9714004b9d2354d8af52f4e15b0f2d1f8

bluebill1049:
Thanks @pmaier983

[...]

PR #469, opened 2020-11-26

pmaier983:
This *should* solve issue #468. 

I had difficulty testing the fix though as currently localhost only saves to localStorage (even if:

```
createStore(
  {
    formData,
    language,
    setting,
  },
  {
    storageType: window.sessionStorage,
  }
)
```

Also I'm not 100% sure this is correct as you mention *user settings* and I interpreted that as `localStorage`, but I could be wrong.

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -4,16 +4,11 @@ import formData from "./src/state/formData"
 import language from "./src/state/language"
 import setting from "./src/state/setting"
 
-createStore(
-  {
-    formData,
-    language,
-    setting,
-  },
-  {
-    storageType: window.localStorage,
-  }
-)
+createStore({
+  formData,
+  language,
+  setting,
+})
 
 export const wrapRootElement = ({ element }) => (
   <StateMachineProvider>{element}</StateMachineProvider>
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -4,11 +4,16 @@ import formData from "./src/state/formData"
 import language from "./src/state/language"
 import setting from "./src/state/setting"
 
-createStore({
-  formData,
-  language,
-  setting,
-})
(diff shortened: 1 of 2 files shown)


bluebill1049:
Thanks, @pmaier983 I think this one is what we after https://stackoverflow.com/questions/56393880/how-do-i-detect-dark-mode-using-javascript

pmaier983:
We have this line in settings.js, it looks like the logic is wrong here. Is this what you mean? 

// settings.js
```
lightMode:
    !window.matchMedia &&
    window.matchMedia("(prefers-color-scheme: dark)").matches,
```

change to 

```
lightMode:
    window.matchMedia &&
    !window.matchMedia("(prefers-color-scheme: dark)").matches,
```


Sorry, I guess I'm confused because we are currently storing users choice in either session storage or local storage, so dark/light mode should persist on reset.

bluebill1049:
> We have this line in settings.js, it looks like the logic is wrong here. Is this what you mean?
> 
> // settings.js
> 
> ```
> lightMode:
>     !window.matchMedia &&
>     window.matchMedia("(prefers-color-scheme: dark)").matches,
> ```
> 
> change to
> 
> ```
> lightMode:
>     window.matchMedia &&
>     !window.matchMedia("(prefers-color-scheme: dark)").matches,
> ```
> 
> Sorry, I guess I'm confused because we are currently storing users choice in either session storage or local storage, so dark/light mode should persist on reset.

oh that's it 👍 i can test it out.

bluebill1049:
awesome! it works!

bluebill1049:
Thank you @pmaier983

bluebill1049 (approved):

pmaier983:
Thanks for explaining the issue! 😄

PR #471, opened 2020-12-01
Changed FormContext to FormProvider in API docs

kasperworks:
Tweaked component name in all languages of V6 docs.


bluebill1049:
oh! thank you. missed that one.

PR #472, opened 2020-12-02

pmaier983:
Make the HTML Template icon dark in light mode.
Fix a bunch of lint warnings 

Screen Shot 2020-12-01 at 10 07 00 PM


bluebill1049 (approved):
Thanks heaps @pmaier983

PR #473, opened 2020-12-02
Note about resetting default values that change

xeoneux:
It might not be entirely clear at first, but this note will help anyone who is struggling with cached default values and would like to change them on the fly :smile:


bluebill1049:
Screen Shot 2020-12-03 at 10 03 29 am

hey, thanks a lot for the PR, maybe we can improve that in the `defaultValues` section? 👍

Issue #470, opened 2020-11-28
codesandbox with each validation rules


bluebill1049:
<img width="960" alt="Screen Shot 2020-11-28 at 11 56 51 am" src="https://user-images.githubusercontent.com/10513364/100490626-cec44700-3170-11eb-8649-74b840fdbde6.png">

I think we are missing (codesandbox)examples of the validation rules, would be great to add them as well.

PR #479, opened 2020-12-08

pmaier983:
This PR:

- adds all but one validation example spoken about in #470. The final example: `validate` I'll finish in the coming week! 
- adds some comments to color defaults and also changes the default from null to false, to make dark-mode default obvious.


bluebill1049:
Awesome stuff! Thanks @pmaier983

PR #485, opened 2020-12-13
Improve resolver examples

oliverbutler:
Make the resolver examples more uniform, and removes some unnecessary code which may add confusion.


bluebill1049:
thanks heaps! 👍

PR #488, opened 2020-12-14
fix: solved file problem not found to configure jest in react native in en

EdnilsonCS:
English documentation update on English test setup


bluebill1049:
Thanks.

PR #489, opened 2020-12-15
add further explanation code about formState props

automatensalat:
Adds a concise code example / do/don't explaining about how to use the `formState` proxy, inspired by [this](https://github.com/react-hook-form/react-hook-form/issues/3673) issue


bluebill1049:
thanks for the update!

PR #494, opened 2021-01-01
Clarify that valueAsNumber and valueAsDate supersede setValueAs

andrewbusch7:
I was using setValueAs alongside valueAsNumber in my code, and expected the setValueAs function to simply receive a number instead of a string. Instead, the function was ignored.

Looking at https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/getFieldValue.ts#L49, I see this experience is by design:

```ts
return valueAsNumber
    ? +value
    : valueAsDate
    ? (ref as HTMLInputElement).valueAsDate
    : setValueAs
    ? setValueAs(value)
    : value;
```

So I thought it would be helpful to clarify this in the documentation.

### DIFF
--- a/src/components/ApiRefTable.tsx
+++ b/src/components/ApiRefTable.tsx
@@ -497,6 +497,9 @@ export default function ApiRefTable({
                 <p>
                   <b className={typographyStyles.note}>Note</b>:{" "}
                   <code>valueAs</code> process is happening after validation.
+                  Also, <code>setValueAs</code> is ignored if either{" "}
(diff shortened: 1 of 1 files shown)


andrewbusch7:
Failure on Lighthouse is https://reactjs.org/docs/error-decoder.html/?invariant=65&args[]=c%C3%B3digo:

```txt
Invalid tag: código
```

I see that text in the codebase inc. directly in some JSX, and I do see that this is the first failure in a while, but I can't reason why my change would have caused that.

bluebill1049:
ty

PR #497, opened 2021-01-04
Fix titles on ts page

unzico:
Hi, thanks for your great library!

I noticed, that the ts page shows the wrong title (and description) for the [UseControllerMethods](https://react-hook-form.com/ts#UseControllerMethods) and [UseControllerOptions](https://react-hook-form.com/ts#UseControllerOptions) sections. Both use the same data as the `UseFieldArrayOptions` section. Probably due to copy and paste.

This PR fixes this.


bluebill1049:
ty very much @unzico

PR #499, opened 2021-01-13
docs(api): update required type accept

neighborhood999:


bluebill1049:
Thank you very much @neighborhood999

src/components/ApiRefTable.tsx
```diff
@@ -139,13 +139,7 @@ export default function ApiRefTable({
...
                 <br />
                 <code className={typographyStyles.typeText}>
-                  {isStandard
-                    ? "boolean"
-                    : `string |
-{
-  value: boolean,
-  message: string | React.ReactElement
-}`}
+                  {`boolean | string | { value: boolean, message: string }`}
```
bluebill1049, line 142: we have a toggler for this, so I will revert this part.

neighborhood999: Sorry my fault, didn't notice the toggler in the above.

PR #502, opened 2021-01-14
Bump Gatsby in hopes that it will fix the api issues

pmaier983:
Not really sure whats causing the 404, so hoping to force a re-deploy of our documentation with some package upgraded.


pmaier983:
Sorry for requesting all of you guys, just want to be sure we can get the documentation up an running ASAP!

jorisre (approved):

PR #506, opened 2021-01-19
Fix form wizard code sandbox

pmaier983:
The bug was brought up here: https://spectrum.chat/react-hook-form/help/official-docs-example-is-not-working-please-check~aa1204e1-d833-4089-b381-c7952d8163d6. I fixed the codesandbox and moved it to our codesandbox "area".


bluebill1049:
thanks for updating that 👍

PR #508, opened 2021-01-23
add-flat-to-required-polyfills

tophergd:
Adds .flat to the list of required polyfills


bluebill1049 (approved):
thank you for including this info in the doc.

PR #512, opened 2021-02-07
useFieldArray default value example fix

Drapegnik:
looks like typo:
Screen Shot 2021-02-08 at 01 10 18


bluebill1049:
100% thanks for the fix.

PR #514, opened 2021-02-15
Add "Handy form validation in React Native" article to resources

dkoprowski:
I've created an article about react-native-hook usage in React Native.

This implementation uses new `useController` hook which makes codebase much better to understand for developers. This is a new approach as most older tutorials describe the usage of `<Controller />` component.


bluebill1049:
Thanks @dkoprowski for the article. ❤️

PR #515, opened 2021-02-17
fixing wrong copy link in getStarted page

StasinskiPL:
npm install yup => npm install @hookform/resolvers yup


bluebill1049:
oh! thanks.

PR #520, opened 2021-03-07
Fix example code

raikusy:


bluebill1049:
thank you very much @raikusy

PR #521, opened 2021-03-08
Update to v7 tag documentation for compatibility with devtools

JackHowa:
I ran into an issue using the wrong version with the v7 with v2 of devtools with `typeError: getValues is not a function` via npm tag https://www.npmjs.com/package/@hookform/devtools

Screen Shot 2021-03-08 at 15 22 13


JackHowa:
Closed in favor https://github.com/react-hook-form/documentation/pull/522 because better branch name and update text

PR #524, opened 2021-03-09
Fix example code

jaeseokk:
* The boolean operator should be fixed to be a valid `disabled` attribute
  * **not** `isDirty` or **not** `isValid`


bluebill1049:
ty 👍

PR #525, opened 2021-03-23
Update formStateUseEffect.ts

Mugilan-Codes:
# Removed TS specific code from JS Example

- removed FormInputs type from line 11. (Since it is a TS specific code in JS)


bluebill1049:
thanks. 👍

PR #531, opened 2021-04-02
Trigger api migration

jorisre:
Add trigger api into the migration page because it's a breaking change.

Screenshot 2021-04-02 at 21 47 31

Related to https://github.com/react-hook-form/react-hook-form/issues/4592


bluebill1049:
thanks, @jorisre that's right. miss it.

PR #532, opened 2021-04-02
Patch: Add new error handling on schema validation

thadeucity:
The code example for the Javascript Schema Validation, had the deprecated way to get the errors from useForm. I just updated it to the V7 way. (Typescript version is ok)


bluebill1049:
thank you.

PR #533, opened 2021-04-03
Make the gallery card clickable

jorisre:
I miss click severals times because only the _Read more_ text is clickable. Now the whole gallery card is clickable

Screenshot 2021-04-03 at 10 26 38


bluebill1049:
awesome! 🔥

PR #539, opened 2021-04-04
Fix migrate code example & add tsconfig.js

raikusy:
* Added tsconfig.json
* Remove @ts-ignore for missing `window.docsearch` typings
* Added global.d.ts to fix `window.docsearch` type.
* Added css.d.ts to fix css types.
* Fixed type in migration guide code example


bluebill1049:
wow, thank you very much for the contribution and help! <3

PR #544, opened 2021-04-07
Fix typos

wdfinch:
This PR fixes additional typos and improves the grammar for the English version of the docs. Note I have updated both `watch` and `getValues` documentation to state that default values are shallowly merged with form values on **each** call. The docs stated this was only true for the first render which is incorrect based on how the library currently works and the RFC. Please let me know if any of my changes result in incorrect documentation.


bluebill1049:
thanks very much @wdfinch

PR #546, opened 2021-04-08
Fix small typo in UseControllerMethods doc

jkulton:
This fixes a small typo I found while reading the docs.

Thanks for all the work on this library and it's docs!


jorisre (approved):
@jkulton Thanks !

PR #549, opened 2021-04-09
Fixed wrong word in typescript data from `Sidemenu`

amantiwari1:
try yourself https://react-hook-form.com/ts and tap `UseFormReturn`  is not working that's why I decided to fix it to avoid error

I simply changed different string and only one variable

So I managed to fix it if there is any mistake please let me know


bluebill1049:
oh wow, thank you for looking into this! and the fix.

PR #552, opened 2021-04-12
Fix forward.ref example

matamatanot:
pass `onBlur`
https://github.com/react-hook-form/react-hook-form/issues/4395#issuecomment-794521966


bluebill1049:
thank you!

PR #564, opened 2021-04-22
error.<field name>.type description is given

kishork2120sparkout:
I can't find description about error.<field name>.type in the given docs, so i have added a brief explanation. kindly ignore if this is mentioned elsewhere.


bluebill1049:
thanks i will update the example as well after this PR is merged.

PR #566, opened 2021-04-24
Update watchCodeTs.ts

soldonii:
`number` field does not exists in `IFormInputs`.
current code invokes type error when following.

I think it should be replaced to `age` field. :)


bluebill1049:
thank you :)

PR #572, opened 2021-05-01
fix `register` code example

alexandr-kim-vl:
Assignment to ref replaced with spread in `setValueAs` and `shouldUnregister` code examples in `register` docs.


bluebill1049:
thank you.

PR #576, opened 2021-05-04
document missing act() warning behavior

automatensalat:
Explaining a (what I found to be a) confusing issue regarding logged warnings about missing `act(...)` calls in unit tests with react-testing-library. 

It's what I took away from these issues: 
https://github.com/react-hook-form/react-hook-form/issues/532
https://github.com/react-hook-form/react-hook-form/issues/4231

Let me know if my interpretation was not correct or you think it should rather be placed into the FAQ section.

### DIFF
--- a/src/components/codeExamples/testingForm.ts
+++ b/src/components/codeExamples/testingForm.ts
@@ -133,37 +133,25 @@ describe("App", () => {
 });
 `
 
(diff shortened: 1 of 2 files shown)


bluebill1049:
thank you very much for this @automatensalat ❤️

PR #578, opened 2021-05-07
Update getStarted.ts

baranberkay96:
I was looking for how it handles errors and I found https://react-hook-form.com/get-started#Handleerrors.

`
    <form onSubmit={handleSubmit(onSubmit)}>
      <Input {...register("firstName", { required: true })} />
      {errors.firstName.type === 'required' && "First name is required"}
      
      <Input {...register("lastName", { required: true })} />
      {errors.lastName && "Last name is required"}
      
      <input type="submit" />
    </form>
`

It gives me this as error. 

`Cannot read property 'type' of undefined`

After the investigation, I found that while component  initializing, errors is empty and trying to get value from nested object could lead us this error. And I want to fix the error on the documentation. I hope it's ok.

### DIFF
--- a/src/components/codeExamples/getStarted.ts
+++ b/src/components/codeExamples/getStarted.ts
@@ -363,7 +363,7 @@ export default function App() {
   return (
     <form onSubmit={handleSubmit(onSubmit)}>
       <Input {...register("firstName", { required: true })} />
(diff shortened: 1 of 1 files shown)


bluebill1049:
thank you

PR #580, opened 2021-05-08
5079 Update little state machine handling on BuilderPage component

leapful:


bluebill1049:
thank you so much @leapful

PR #586, opened 2021-05-11
Improve get started example code types

bdmason:
I have `"strict": true` in my `tsconfig.json`, when I pasted in the first code sample from `/get-started` the lack of types for `onSubmit` was highlighted as an error.

When looking into the code for the docs I noticed that whilst a lot of the examples weren't implicitly setting `data` in `onSubmit` as `any`, they weren't using the more thorough `SubmitHandler` type.

I also spotted that `uiLibraryHookInput` had a stray type in the plain JS example, so I removed that.


bluebill1049:
thank you mate!

PR #591, opened 2021-05-15
docs: update `getValues` documentation

imballinst:
Thanks for the awesome documentation! I can't ask for more, it has proper explanation for every field, good description, CodeSandbox examples, even TypeScript. However, there was one thing that made me confused. 

## Before

This part below, the docs (https://react-hook-form.com/api/useform/getvalues) says when we supply the `firstName`, it returns `lastName` and vice versa. I think it should be consistent with the argument that we pass to the function, as reflected by the Sandbox in the same section.

image

## Checklists

- [x] I have tested the documentation using `yarn start`
- [x] I have run `yarn format`

## After

This PR swaps the `.firstName` and `.lastName` in the documentation. I have cross-checked this with the sandbox in https://codesandbox.io/s/react-hook-form-v7-getvalues-2eioh and it worked as expected—`getValues(field)` will return `{ [field]: fieldValue }`.

image

### DIFF
--- a/src/data/en/api.tsx
+++ b/src/data/en/api.tsx
@@ -1503,7 +1503,7 @@ setValue('notRegisteredInput', { test: '1', test2: '2' }); // ✅
                   <code>getValues("yourDetails.firstName")</code>
                 </td>
                 <td>
-                  <code>{`{ lastName: '' }`}</code>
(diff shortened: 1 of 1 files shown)


bluebill1049:
thank you very much for the fix!

PR #593, opened 2021-05-17
docs: add small note about `submitFocusError`

devrnt:
I noticed that the autofocus on error doesn't work when using "custom" components like mentioned in the documentation (`submitFocusError`). However, forwarding the ref makes this work. I thought it was worth mentioning

```jsx
const Input = React.forwardRef((props, ref) => {
  return <input {...props} ref={ref}></input>;
});
```


bluebill1049:
thank you @devrnt ❤️

PR #599, opened 2021-05-28
Note valueAs breaking changes in migration guide

ivantm:
👋 Hello and thanks for react-hook-form!

I'm knee deep in a migration from v6 to v7 and have been caught by https://github.com/react-hook-form/react-hook-form/issues/4673 which wasn't mentioned in the migration guide, so I added it here in case it helps anyone in the future.

There are a bunch of other breaking changes I've noticed in the typings (e.g. FieldArray* -> ArrayField*), but I've been kicking that can down the road for now 😄


bluebill1049:
thank you very much!

PR #601, opened 2021-06-01
fix: reference the code example to v6 in ApiPageV6

andykao1213:
I found that the [sample code of reset](https://react-hook-form.com/v6/api#reset) is misplaced with v7 (`keepErrors` is v7. In v6 it was called `error`). Then after tracing the source code, I found that there are lots of code examples in v6 that are referencing v7. This PR aims to reference the correct code examples. Thank you!


bluebill1049:
thanks, buddy!

PR #614, opened 2021-06-12
Fix incorrect variable names

dogukanarslan:
There are remaining `touched` keywords which is used in v6 and replaced with `touchedFields` in v7. Also in code examples formState is being destructured to touched by `formState: { touched }`  and then trying to reach the `formState`. It is not possible after destructuring. This PR fixes those issues.


bluebill1049:
thanks a lot.

PR #616, opened 2021-06-14
Fix usecontroller docs mistake

alexandr-kim-vl:
Remove `value` from onBlur event in useController docs to match corresponding type:
```ts
export declare type ControllerRenderProps<TFieldValues extends FieldValues = FieldValues, TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>> = {
    onChange: (...event: any[]) => void;
    onBlur: () => void;
    value: FieldPathValue<TFieldValues, TName>;
    name: TName;
    ref: React.Ref<any>;
};

`


bluebill1049:
thanks!

PR #618, opened 2021-06-16
Fix Scroll Behaviour

### LINKED ISSUE #617: Scroll To Top Button is Not Working
# Description

Scroll To Top Button is not working on Firefox Developer Edition.

As soon as the button is clicked, it hides, and the click operation is never completed.

SNikhill:
# Description

Turns out that on focus, the button was set to `display: none` and hence, as soon as the user used to click, the button's focus state turned active and it was removed from DOM, therefore, leaving the click operation _"hanging"_.

## Other Information

- I tested the change in Firefox Developer Edition + Brave Browser.
- I have changed the cursor to `pointer` to increase interactivity.
- fixes #617

### DIFF
--- a/src/components/layout.css
+++ b/src/components/layout.css
@@ -355,12 +355,12 @@ table {
 .scrollToTop,
 .scrollToTop:focus {
   border: 1px solid var(--color-light-grey);
-  display: none;
 }
 
 .editPage:hover,
 .scrollToTop:hover {
   border-color: var(--color-secondary);
+  cursor: pointer;
 }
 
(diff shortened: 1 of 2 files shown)


SNikhill:
Smooth behaviour would be better, right?

bluebill1049:
> Smooth behaviour would be better, right?

yes. thanks for the fix 👍

SNikhill:
> > Smooth behaviour would be better, right?
> 
> yes. thanks for the fix +1

Smooth it is. Thank you for this project. :)

PR #619, opened 2021-06-20
Import v6 code examples from v6 folder

dogukanarslan:
In V6 API, controller section code examples are exported from v7 code examples. This PR fixes those imports.


bluebill1049:
Thank you!

PR #621, opened 2021-06-23
Fix ReactNative part of getStarted page.

YutaSugimura:
- Removed duplicate props.
- The argument of onChange was duplicated with value, so it was fixed.


bluebill1049:
thank you.

PR #626, opened 2021-06-29
docs(devtool): stable devtool

neighborhood999:
## Description
Changed install dependency version, because current will install the beta version and it still has some issues.

## Reference
https://github.com/react-hook-form/devtools/issues/79

Signed-off-by: Jie Peng <im@jiepeng.me>


bluebill1049:
thank you very much @neighborhood999 ❤️

PR #630, opened 2021-07-13
update migration guide (en) to include shouldUnregister

ZachMayry:
I wasn't sure on the ordering so I just put it on the bottom.
Open to any suggestions on verbiage


bluebill1049:
thank you!

PR #638, opened 2021-07-29
Install @hookform/devtools as a dev dependency

imadatyatalah:


bluebill1049:
Thanks for the PR, it's fine to keep as dep. Devtool actually gets omit during prod build.

PR #639, opened 2021-07-30
Fix typo in migration v6 to 7 page

atobaum:
There was no space between plain text and code in migration guide
https://react-hook-form.com/migrate-v6-to-v7/


bluebill1049:
thanks

PR #641, opened 2021-08-02
Fix typo in the modal question of FAQ

rayrw:
In FAQ's `How to work with modal or tab forms?` 's `Custom Register` tab,
I think the parameter order of calling `handleChange` in the `lastName` field is reversed.


bluebill1049:
thank you!

PR #649, opened 2021-08-17
Add more information about the callback version of watch

joshkel:
Specific changes:

- Add more specific summary text.  For languages other than English, I simply duplicated the previous summary text, since I don't have a way to translate into those languages.
- Add an example.
- Add a specific note and example about cleaning up / removing the subscription; I'd like confirmation that I'm doing this correctly.
- Fix the return type of the callback version.


bluebill1049:
thanks for the update.

PR #653, opened 2021-08-30
React.RefObject -> React.Ref

pmaier983:
Fixes Issue #647. React.Ref is what core uses, and it is just a wrapper around React.RefObject that also includes callback refs & null:
reactRefExample

### DIFF
--- a/src/components/ApiErrors.tsx
+++ b/src/components/ApiErrors.tsx
@@ -87,7 +87,7 @@ export default React.memo(
                 </td>
                 <td>
                   <span className={typographyStyles.typeText}>
-                    React.RefObject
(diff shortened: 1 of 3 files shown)


bluebill1049:
Thanks! @pmaier983

bluebill1049 (approved):

jtiala:
I don't think this is correct. RHF typings clearly say `ref: RefCallBack;` (controller.d.ts:15), which is a void function `export declare type RefCallBack = (instance: any) => void;`(form.d.ts:90). It can never be a `RefObject` or `null` if I'm correct. @bluebill1049 ?

PR #665, opened 2021-09-12
Expand TS -> Typescript

pmaier983:
This will allow Algolia Search to better identify the TS Support Page

*FYI: I tested the change at several different screen sizes and it still formats well


bluebill1049 (approved):
Awesome. thanks

PR #667, opened 2021-09-17
Update example to call with proper argument

johnathanludwig:
Using `useWatch` like the previous example showed would instead provide all the values for all fields.


bluebill1049:
thank you! oops even i don't know the API that well hahahah

PR #668, opened 2021-09-19
7.16.0

bluebill1049:
doc update for 7.16.0


michalbundyra:
Changes looks good, but maybe also worth updating migration guide - for `register` https://react-hook-form.com/migrate-v6-to-v7/ ?

bluebill1049:
@michalbundyra thanks for the suggestion, will do probably later. 👍

PR #674, opened 2021-10-04
Added a more visible search

getTobiasNielsen:
Based on this userfeed back the search wasn't visible enough

sdsdsdsdsd - Copy

The search is now open on large monitors, and automatically becomes smaller when the screen size is made smaller, also keeps the darkmode and version + language options open on large monitors.


bluebill1049:
awesome stuff! thanks @getTobiasNielsen 🙏

PR #678, opened 2021-10-15
Add `disabled` props to useWatch props signature

GaelS:
First of all: thanks for this great lib ! 

I catch what I think might be an error in the documentation : according to react-hook-form source code, `useWatch` takes the following object as argument:

```
export type UseWatchProps<TFieldValues extends FieldValues = FieldValues> = {
  defaultValue?: unknown;
  disabled?: boolean;
  name?:
    | FieldPath<TFieldValues>
    | FieldPath<TFieldValues>[]
    | readonly FieldPath<TFieldValues>[];
  control?: Control<TFieldValues>;
};
```
but `disabled` was missing and mispelled in the according documentation.

Hope this change is relevant.

Cheers

### DIFF
--- a/src/components/UseWatchContent.tsx
+++ b/src/components/UseWatchContent.tsx
@@ -22,7 +22,7 @@ export default function UseFieldArray({
           useWatch:{" "}
           <span
             className={typographyStyles.typeText}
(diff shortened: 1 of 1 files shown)


GaelS:
not sure I can see my updates in the preview :thinking:

PR #682, opened 2021-10-21
fix: update wrong name in watch page

nvh95:
In this file, we have 
```
    name: 'firstName', // without supply name will watch the entire form, or ['firstName', 'lastName'] to watch both
```
However, the input name is `last` instead of `lastName`. I believe this is a typo.


bluebill1049:
thank you.

PR #689, opened 2021-11-02
Remove outdated RN Hermes polyfill advice

GfxKai:
Proxy has been removed from codebase, so react native will share the same perf enhancement as web. Regardless, Hermes v0.7+ supports Proxy by default.


bluebill1049:
thanks!

PR #699, opened 2021-11-24
docs: fixed missing render props in multiple ErrorMessage examples

getTobiasNielsen:
Assume I made a mistake, because the examples are inside a string literal.


bluebill1049:
Thanks!

PR #703, opened 2021-12-05
feat(form builder): added password type to form builder

akhil-gautam:
Screenshot 2021-12-05 at 1 33 37 PM


bluebill1049:
thanks!

PR #705, opened 2021-12-09
Added register type - Update getStarted.tsx

ritikbanger:
Added register type for typescript users. The type is asked by community in so many questions in discussions.


bluebill1049:
awesome! thanks @ritikbanger

PR #708, opened 2021-12-10
[docs] `register()` wording clarity

gregfenton:
"It" is ambiguous in this context.  Replacing with the explicit field value being discussed (`name`)


bluebill1049:
thanks.

PR #718, opened 2022-01-22
Fix incorrect property name for fieldState.error

knowler:
Seems like this was copied from the previous item (`invalid`).


bluebill1049:
oh, thank you! sorry miss the PR.

PR #719, opened 2022-01-24
fix: added the definition of handleClick

mrsekut:
In the code example of [resetField](https://react-hook-form.com/api/useform/resetfield), the definition of handleClick was missing, so I added it.
Fixed to match the code in [CDB](https://codesandbox.io/s/resetfield-with-options-iw4wd).


bluebill1049:
thank you 🙏

PR #730, opened 2022-02-10
📝 docs: add shouldSelect focus option

Moshyfawn:
react-hook-form/react-hook-form#7781
react-hook-form/react-hook-form#7318

### DIFF
--- a/src/components/useForm/SetFocus.tsx
+++ b/src/components/useForm/SetFocus.tsx
@@ -26,7 +26,7 @@ export default ({ currentLanguage }) => {
               <h2>
                 setFocus:
                 <span className={typographyStyles.typeText}>
-                  {`(name: string) => void`}
+                  {`(name: string, options: SetFocusOptions) => void`}
                 </span>
               </h2>
             </code>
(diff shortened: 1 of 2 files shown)


bluebill1049:
nice! I was going to do it.

bluebill1049:
would you like to add a quick example in the code example as well?

Moshyfawn:
> would you like to add a quick example in the code example as well?

What do you think about a code snippet under the `options.shouldSelect` description? It doesn't feel like that big of a code difference from the basic example to do it's own sandbox, in-app usage example and it's TS version

image

bluebill1049:
Sounds good to me with the example. Single paragraphs probably don't need a list, we can start to introduce a list when we have multiple dot points.

Moshyfawn:
> Single paragraphs probably don't need a list, we can start to introduce a list when we have multiple dot points.

Good point

bluebill1049:
Just one minor, for consistency could you move rules under the Props which I have actioned based on your feedback. Let's roll this feature out this weekend as well ☺️

Moshyfawn:
> Let's roll this feature out this weekend as well ☺️

I saw that, yeah!
I'm not a fan of having bright heading for a section as the section title seems like the only "important" information to me and my break skip passes the rest of the info. Maybe it's just me, but I'll much more reliably read the rest of the section, if it's all highlighted with a slightly different background along with the heading.
Not saying we should do that instead, just my personal feedback ;)

bluebill1049:
@Moshyfawn Thanks for the feedback, I change the title underline because there are two many pink headers.

PR #741, opened 2022-02-25
Add description of undefined onChange to useController documentation

david-morris:
Explicitly state that calling `onChange` with `undefined` will clear the named field when using `useController`.

This is a useful operation, since some types of form inputs, such as number input, can be cleared in a way which means something different than any allowable value.

### DIFF
--- a/src/data/en/api.tsx
+++ b/src/data/en/api.tsx
@@ -2151,7 +2151,9 @@ setValue('notRegisteredInput', { test: '1', test2: '2' }); // ✅ sugar syntax t
               </li>
               <li>
                 <p>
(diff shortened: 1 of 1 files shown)


bluebill1049:
> It should be assigned to the onChange prop of the input and value should not be undefined.

thanks for the PR, i would avoid `undefined` at all with `useController`, it may lead to unexpected behavior, depend on your controlled input type, you should either use `null` or empty `string` base on your input's default state value

david-morris:
@bluebill1049 thanks for that info!  How would this sound?
> Avoid calling `onChange(undefined)`. This results in undefined behavior which may change by the type of your controlled input and minor or patch versions.

That way, the issue doesn't come up on GitHub again.

bluebill1049:
thanks for the PR.

PR #743, opened 2022-02-27
Note about FormProvider performance when using RHF Devtools

cgarrovillo:
After hours of head-scratching and trying to figure out why my forms are not performing the way they should, and spending hours on optimizations (ie. memoization,  consolidating input fields),  I came to find out that the source of my bottleneck was RHF Devtools.

This PR adds this note just below the FormProvider performance description in Advanced Usage,  in hopes it will save others time, headaches, engineering hours & money that is usually the cost of doing performance optimizations.

### DIFF
--- a/src/data/en/advanced.tsx
+++ b/src/data/en/advanced.tsx
@@ -286,23 +286,34 @@ export default {
   formContext: {
     title: "FormProvider Performance",
     description: (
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for the additional context.

PR #751, opened 2022-03-24
Mention keyName prop for useFieldArray

kilya11:


bluebill1049:
Thanks for the PR, this props is no longer required and will be deprecated in this next version.

Mini-ghost:
@bluebill1049 Why `keyName` is no longer required? Has any discussion mentioned this?

bluebill1049:
@Mini-ghost it's not a breaking change, in fact, `keyName` prop is deprecated in the next major version, useFieldArray key name will no long conflict with your actual form data.

PR #755, opened 2022-03-31
Use correct prop name isTouched instead of isTouch

sebastianlp:
Currently we use `isTouch` to refer this property but the correct one should be `isTouched`


bluebill1049:
oops, thanks!

bluebill1049 (approved):

PR #760, opened 2022-04-07
mention that control is optional in useFormState

ms10596:
mention that control is optional in useFormState if you're using FormContext.


bluebill1049:
thank you.

PR #766, opened 2022-04-12
Fix EN grammatical error in ApiGallery

rmcauley:
There is a grammatical error in the English ApiGallery file.


bluebill1049:
thanks

PR #772, opened 2022-05-13
Update api.tsx

leanjunio:
The method should be `unregistered`


bluebill1049:
thanks for the PR, but the doc is correct with using register API for input

leanjunio:
Ah gotcha. Thanks for the clarification @bluebill1049

PR #784, opened 2022-06-16
fix: incorrect resource link for - Using React Hook Form with Materia…

thevipinmishra:
Found an incorrect link in the resources section.

Link was incorrect for - **Using React Hook Form with Material-UI Components**


bluebill1049:
thanks

PR #788, opened 2022-06-24
docs(example): add destructure on useForm example

bmuenzenmeyer:
Hello!

I was following along with the docs (excellent, btw!) and noticed a missing destructure in the `Submit with Reset` code example at https://react-hook-form.com/api/useform/reset

image

Using the codesandbox feature, and copy-pasting ***this exact code***, yields an error consistent with my local environment:

image

Using the codesandbox feature and not changing the initial code that loads in the sandbox, you'll notice the additional destructure present https://codesandbox.io/s/react-hook-form-handlesubmit-with-reset-xu1zu

image

If I am missing some quirk of the architecture or your process - feel free to close

### DIFF
--- a/src/components/codeExamples/submitReset.ts
+++ b/src/components/codeExamples/submitReset.ts
@@ -6,6 +6,7 @@ function App() {
     register,
     handleSubmit,
     reset,
+    formState,
     formState: { isSubmitSuccessful }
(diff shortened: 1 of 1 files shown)


bluebill1049:
thank you for the fix.

PR #789, opened 2022-06-25
Add: onSubmit method and required with message example

zorrolisto:
Its very nice to have a library like this one, hope this help a little bit with the getStarted page.

Added:
1.  Method onSubmit
2. Example for required with message


bluebill1049:
thanks 🙏

PR #790, opened 2022-07-04
Update UnRegister.tsx

TawalMc:
The documentation mention keepValid instead of keepIsValid for the unregister options, so I fixed it by putting the right option name (keepIsValid) in the code examples and the description.


bluebill1049:
thank you.

TawalMc:
> thank you.

I beg you

PR #796, opened 2022-07-22
useForm API: Improve correctness of JS terminology

buondevid:
I think that `useForm()` accepts only **one** optional argument and the things listed in the example are **its properties**.


bluebill1049:
thanks 🙏

PR #798, opened 2022-07-27
fix: uppercase "Input" instead of "input"

andremralves:
I guess the code example in section "Integrating with global state" was supposed to use "input" (all lowercase) instead of "Input" component. Because there is no "Input" component being defined.


bluebill1049:
thanks

PR #800, opened 2022-07-29
Fix v6 watch API code example

adcdr:
The v6 documentation currently shows an example for a callback version of watch, which is only available in v7:

```
  // Callback version of watch.  It's your responsibility to unsubscribe when done.
  React.useEffect(() => {
    const subscription = watch((value, { name, type }) => console.log(value, name, type));
    return () => subscription.unsubscribe();
  }, [watch]);
```
https://react-hook-form.com/v6/api#watch

This seems to have happened because the imports for the code examples point to the latest examples directory, rather than the v6/ directory.

The same issue was mentioned by another user [here](https://github.com/react-hook-form/react-hook-form/issues/1905#issuecomment-928716607)

### DIFF
--- a/src/components/v6/ApiWatchV6.tsx
+++ b/src/components/v6/ApiWatchV6.tsx
@@ -1,13 +1,13 @@
 import * as React from "react"
 import CodeArea from "../CodeArea"
-import watchCode from "../codeExamples/watchCode"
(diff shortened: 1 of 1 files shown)


bluebill1049:
that's right. thank you!

PR #804, opened 2022-08-03
improve doc on getFieldState

bluebill1049:
close issue: https://github.com/react-hook-form/react-hook-form/issues/8786#issuecomment-1204570567

### DIFF
--- a/src/components/useForm/GetFieldState.tsx
+++ b/src/components/useForm/GetFieldState.tsx
@@ -73,16 +73,16 @@ export const GetFieldState = () => {
                       <p>
                         This is an optional prop, which only required to supply
                         if formState is not been read/subscribed from the
-                        useForm hook.
+                        <code>useForm</code> or <code>useFormState</code> hook.
                       </p>
                       <CodeArea
                         rawData={`// dirtyFields are subscribed
 const { formState: { dirtyFields } } = useForm()
-getFieldState('firstName') // function aware of dirty fields subscription and return form state accordingly
+getFieldState('firstName') // dirty fields subscription is done and return form state accordingly
 
 // formState is not read/subscribed
 const methods = useForm();
-const { error } = getFieldState('firstName', methods.formState) // provide formState as optional argument
+const { error } = getFieldState('firstName', methods.formState) // provide formState as optional argument due to lack of subscription detail
 `}
                       />
                     </td>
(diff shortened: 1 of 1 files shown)


src/components/useForm/GetFieldState.tsx
```diff
@@ -175,18 +188,36 @@ getFieldState('non-existent-name'); // ❌ will return state as false and error
                 />
               </li>
               <li>
-                <p>formState will need to be subscribed.</p>
+                <p>
+                  <code>formState</code> from <code>useForm</code> will need to
```
fzaninotto, line 192: I don't understand this sentence. Do you mean

> `getFieldState` works by "subscribing" to the form state. That's why it requires the form state as second argument.
> ```jsx
> const { error } = getFieldState('firstName', methods.formState)
> ```
>
> Other hooks (`useForm` or `useFormState`) also "subscribe" to the form state. For instance, this call subscribes to the `errors` part of the form state:
> ```jsx
> const { formState: { errors} } = useForm() // errors is reactive because useForm placed a subscription on it
> ```
> If such a subscription already exists, you can omit the second parameter to `getFieldState`, as it will take advantage of the existing subscription. 
> ```jsx
> const { error } = getFieldState('firstName');
> ```

Writing this, I realize that there may be a correspondence between the form subscriptions (e.g. `errors`)  and the field subscriptions (e.g. `error`). You should probably explain that correspondence somewhere.

src/components/useForm/GetFieldState.tsx
```diff
@@ -175,18 +188,36 @@ getFieldState('non-existent-name'); // ❌ will return state as false and error
...
                 <p>You can subscribe to the formState in two different ways:</p>
 
                 <ol>
                   <li>
                     <p>
-                      You can simply define the required state information when
-                      destructuring the returned value of a relevant hook, such
-                      as <code>useForm</code> (at global level) or{" "}
-                      <code>useFormState</code> (at hook level).
+                      You can define form state subscription at the{" "}
```
fzaninotto, line 203: This is really hard to follow. I really wonder if it is necessary to explain this use case? Why not only document the use case where the developer has to pass the form state as second parameter?

bluebill1049:
thanks for the comments @fzaninotto they are all valid, I will simplify the doc a bit so it's not misleading the developers. The key information here is `getFieldState` is just a function that needs a `formState` (dirty/errors/touched) subscription before it works properly.

PR #812, opened 2022-08-31
Rules text turns black in light mode. fix #811 

### LINKED ISSUE #811: When in light mode, the text color of `Rules` in useFormContext is the same as the background and is unreadable.
screenshot 2022-08-31 19 12 04
https://react-hook-form.com/v6/api#useFormContext

When the `Rules' text in useFormContext is in light mode, as shown in the image, it is unreadable because it is white, the same color as the background.

mochi-sann:
fix #811
screenshot 2022-08-31 19 53 05

The word Rules is now black in light mode, as shown in the image

### DIFF
--- a/src/styles/typography.module.css
+++ b/src/styles/typography.module.css
@@ -54,7 +54,7 @@ h3.subTitle:before {
 }
 
 .rulesTitle {
-  color: white;
+  color: currentColor;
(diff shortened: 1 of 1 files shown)


bluebill1049:
thank you for the fix! 🙏

PR #819, opened 2022-09-10
Fix onSubmit `data` type on useForm zod resolver code example

taufiqfebriant:
Replacing `IFormInput` with `Schema` on `onSubmit` function


bluebill1049 (approved):
thanks

PR #833, opened 2022-10-01
include missing prop to forwardRef example

yami03:
image

[Documentation link](https://react-hook-form.com/api/useform/register/#example:~:text=How%20to%20work%20with%20innerRef%2C%20inputRef%3F)

I added the name prop. 

I didn't add a name prop to my code like a document at the first time.
My code didn't validate well when the mode was `all`, `onChange`, and `onBulr`. 
I found the name prop was the cause. 

I think it would be better to add a name prop to the document.


bluebill1049:
thanks!

PR #836, opened 2022-10-05
feat: add `UseFormReturn` type

### LINKED ISSUE #835: Add/Replace TypeScript types for `UseFormReturn` section on `/ts` page
## Overview

Currently `UseFormReturn` section is included examples about how to use the type, but there's no detailed type field for`UseFormReturn` itself. To make it easier for users to see what the `UseFormReturn` type is, I think we can add the type and persist the code examples like this:

image

What do you think? I can help add the type if this is possible

abdmmar:
Close #835

### DIFF
--- a/src/data/ts.tsx
+++ b/src/data/ts.tsx
@@ -190,9 +190,32 @@ export default function App() {
   UseFormReturnRef: {
     title: "UseFormReturn",
     description: (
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
LGTM thanks!

PR #842, opened 2022-10-08
fix: edit links for api and advanced pages

griimick:
Specifically targeted only data sources under V6 and V5 folders for legacy-docs.

Points to note:
1. Modified source prefix to point to `V6-V5` branch instead of `master` as legacy docs are served from that branch.
2. Noticed there are two versions of `/advanced-usage`. For now handled the links as they are shown in docs.
    - `V7, *` https://github.com/react-hook-form/documentation/blob/v6-v5/src/data/en/advanced.tsx
    - `V6, en` https://github.com/react-hook-form/documentation/blob/v6-v5/src/data/V6/en/advanced.tsx

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -22,6 +22,7 @@ const Layout = (props: {
   const { currentLanguage } =
     language && language.currentLanguage ? language : { currentLanguage: "en" }
   const lightMode = state?.setting?.lightMode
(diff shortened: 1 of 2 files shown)


bluebill1049:
thank you very much

griimick:
Thank you for this project as well :)

PR #848, opened 2022-10-13
FIX: Add missing reference to `remove` in the `useFieldArray` example

manassra:
The code example referenced in the [`useFieldArray` documentation page](https://react-hook-form.com/api/usefieldarray) is missing to define the `remove` that is referenced below (see screenshot).

image


bluebill1049:
thank you!

Issue #838, opened 2022-10-07
[Docs] Save user theme preference


michaeloliverx:
**Is your feature request related to a problem? Please describe.**
I use the docs of `react-hook-form` quite often and I always have to switch the theme back to light mode. 

**Describe the solution you'd like**
I would love if the website remembered my previous selection.

**Describe alternatives you've considered**
Manually doing it each time 😢 

**Additional context**

bluebill1049:
Yea sure, PR welcome on this 👍

PR #849, opened 2022-10-15
#838 Apply storage type "window.localStorage" on little state machine

leapful:
Start using "window.localStorage" for little state machine in order to keep theme settings on multiple tabs.

Related issue: https://github.com/react-hook-form/documentation/issues/838


bluebill1049 (approved):
Thanks it works. thank you @leapful

PR #850, opened 2022-10-15
Update DevTools.tsx | Add -D flag to install cmd

bdugg9119:
Added the `-D` flag to the installation command so users can streamline copy-pasting the command without needing to remember to add the flag when running the command.


bluebill1049:
thanks

PR #852, opened 2022-10-17
Proposal change to the copied code

AdaltonLeite:
## Description 

We are currently opening an `alert` to let you know that the code has been copied. My proposal is to improve the user experience by informing on the button itself that the code is already copied in its transfer area. If this change makes sense, we create this action for the others copy buttons. 

## BEFORE
ezgif-3-461f27a320

## AFTER
ezgif com-gif-maker (1)

### DIFF
--- a/src/components/ClipBoard.tsx
+++ b/src/components/ClipBoard.tsx
@@ -1,0 +1,39 @@
+import React, { useState } from "react"
+import generic from "../data/generic"
+
+const ClipBoard = ({
+  currentLanguage,
+  className,
+  onClick,
+}: {
+  onClick: () => void
+  className?: string
+  currentLanguage: string
+}) => {
+  const [copiedCode, setCopiedCode] = useState<boolean>(false)
+
+  React.useEffect(() => {
+    if (copiedCode) {
+      setTimeout(() => {
+        setCopiedCode(false)
+      }, 3000)
+    }
+  }, [copiedCode])
+
+  return (
+    <button
+      className={className}
+      onClick={() => {
+        onClick()
+        setCopiedCode(true)
+      }}
+      aria-label={generic.copied[currentLanguage]}
+    >
+      {copiedCode
+        ? generic.codeCopied[currentLanguage]
+        : generic.copy[currentLanguage]}
+    </button>
+  )
+}
+
+export default ClipBoard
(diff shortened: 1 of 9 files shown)


bluebill1049:
that looks good, would you like to apply to all so it's consistent? perhaps a custom hook for the job.

AdaltonLeite:
Yeah. Make sense create for all. I'll change the PR.

AdaltonLeite:
@bluebill1049 I created a component `ClipBoard`, the function of this component is just render a text and change a label with **COPY** and **COPIED**

I change on this pages/components: 

- [x] Get Started
ezgif com-gif-maker (2)

- [x] Dev Tools
ezgif com-gif-maker (3)

- [x] Builder page
ezgif com-gif-maker (4)

- [x] CodeArea
ezgif com-gif-maker (5)


I searched the whole project and couldn't find any more alerts modal with clipboard.

bluebill1049:
thanks a lot for making this happend. 👍

AdaltonLeite:
Thanks you for this beautiful library!

src/components/ClipBoard.tsx
```diff
@@ -0,0 +1,39 @@
...
+const ClipBoard = ({
+  currentLanguage,
+  className,
+  onClick,
+}: {
+  onClick: () => void
+  className?: string
+  currentLanguage: string
+}) => {
+  const [copiedCode, setCopiedCode] = useState<boolean>(false)
```
bluebill1049, line 13: worth adding a ref with setTimeout in case of a memory leak.

```tsx
const timer = useRef()

React.useEffect(() => () => clearTime(timer.current))
```

bluebill1049 (commented):
Looks great, just one minor feedback.

src/components/ClipBoard.tsx
bluebill1049: I will address this tho, thanks for the awesome work!

bluebill1049: done :) https://github.com/react-hook-form/documentation/commit/a33698125936cc290d91356ca109843e331f1505

AdaltonLeite: Excellent point! Thanks so much.

PR #859, opened 2022-10-26
fix typo kay->key for list

programistka:
Fix for ESLint error - `li` element had `kay` property instead of `key`


bluebill1049:
Haha thanks for that.

PR #860, opened 2022-10-26
add yarn version of installing commands

programistka:
Add `yarn` command version for installing the package.


bluebill1049:
Thanks a lot for the PR, I think we will keep a single NPM for now considering there is an even better option such as pnpm. This is not a discourage for PRs, in fact, please keep them coming and I appreciate each one of them.

programistka:
> Thanks a lot for the PR, I think we will keep a single NPM for now considering there is an even better option such as pnpm. This is not a discourage for PRs, in fact, please keep them coming and I appreciate each one of them.

Sure thing, thank you! 🚀

PR #866, opened 2022-10-27
Enhancement: Input radio validade on page builder

AdaltonLeite:
I add a required validate on input **radio** and **select** to solve this issue: https://github.com/react-hook-form/documentation/issues/834. 

I think we don't have a bug on this page if you add an `Options` on the page build it will generate a code for you, but if you don't add the `Options` they don't create a code for you. My proposal is required the options field. Looks good for you? 

ezgif com-gif-maker (7)


bluebill1049 (approved):
thank you again 🙏

PR #867, opened 2022-10-28
Fix: Add redirect to legacy api doc

AdaltonLeite:
## Description

Today on the main site(https://react-hook-form.com/) we have a bug when the user access the API documentation and change the **version to 6 or 5**:
ezgif com-gif-maker (8)

This bug occurs because we don't have a page folder to v6 and v5 api: 
Captura de Tela 2022-10-28 às 16 27 31

My proposal is to redirect the user to legacy documentation when they change the version v6(https://legacy.react-hook-form.com/v6/api) and v5(https://legacy.react-hook-form.com/v5/api) documentation. On the legacy we have a page for this versions:
Captura de Tela 2022-10-28 às 16 29 10

Recording the issue fix. 
ezgif com-gif-maker (9)

Look good for you? What do you think?

### DIFF
--- a/src/components/ApiGallery.tsx
+++ b/src/components/ApiGallery.tsx
@@ -36,16 +36,20 @@ export default function ApiGallery({ defaultLang }) {
   const onChange = (e) => {
     const version = parseInt(e.target.value)
 
-    navigate(`/v${version}/api/`)
-
     actions.updateSetting({
       version,
     })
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
thank you! you are doing a better maintain job than i do. hahaha ❤️

PR #868, opened 2022-10-28
Proposal: redirect button "edit on github" to a correct page

AdaltonLeite:
Hi, I'm here again! 😊

I proposal a logic to redirect to a correct page when user click on edit button on API pages, today they redirect to api.tsx generic page. (https://github.com/react-hook-form/documentation/edit/master/src/data/en/api.tsx)

My proposal is to redirect to a correct github page with user access on the doc. 

Obs: I create a logic to remove the last "/" if the user access urls to end with ** /**, this break a redirect link to github: 

Example:
- [x] https://react-hook-form.com/api/
- [x] https://react-hook-form.com/api/useform/resetfield/

I think this PR solve this issue: https://github.com/react-hook-form/documentation/issues/831


bluebill1049 (approved):
thanks!

PR #889, opened 2023-01-04
Documentation copy improvement recommendations

sethlewandowski:
I have been using RHF professionally at the agency that I work for. It has been great to work with! I noticed a few areas in the documentation that could be worded differently to improve understanding/clarity. 

I did not attempt to change any of the meaning, just recommendations on wording that helps clarify what is being communicated.


bluebill1049 (approved):
Thanks a lot for this! It's an awesome improvement.

PR #891, opened 2023-01-12
Add forgotten control in code example

AliRoshaniDev:
This is simple and clear. The control prop is not given to useWatch.


bluebill1049:
thanks! In case you didn't know, it is optional if you are using the FormContext api.

PR #893, opened 2023-01-18
Fixed undefined error on builder

shmohit2002:
The BuilderPage had an issue when we create new form fields the code doesn't generates for the validation and type also gets as "undefined" when we create 2 fields in a row.
In the below example I tried to create two field with basic and same configuration without and validation. The first one with the field name "test1" got successful and showed up as per the inputs but for the second input with field name "test2" was created but with undefined as the type of the input as shown in the below picture
image

### DIFF
--- a/src/components/BuilderPage.tsx
+++ b/src/components/BuilderPage.tsx
@@ -212,7 +212,7 @@ function BuilderPage({
           <select
             aria-label="Select type"
             name="type"
-            ref={register}
+            ref={register({required: false})}
             defaultValue={editFormData.type}
           >
             <option value="text">Text</option>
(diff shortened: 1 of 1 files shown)


shmohit2002:
I have tested the code after the changes made and the following output is generated. Hence the code is now working for any input type and no undefined error is occurred.
image

bluebill1049:
Thanks for the patch, the name input is actually required which is the root cause in the first place. Maybe we should disable other input until name field has value.

shmohit2002:
Yeah, the name input is required and the fix is made for all the other input fields which was causing trouble. Hence, I have tweaked the code for all the other fields except for "name" input field. And the same is tested and mentioned in the image of the previous comments.

bluebill1049 (approved):
Thanks for the patch.

PR #899, opened 2023-01-22
Fixed issue with validation checkbox

shmohit2002:
Validation checkbox issue whenever used the validation popup isn't hidden by default from the user.

Test performed:
The following was submitted for the first time
image
and after clicking on create the validation popup (which contains required, min, max fields) continued to show up while the checkbox was unchecked
image

Modified: When clicking on create the state of popup was changed to make sure the popup is hidden


bluebill1049 (approved):
thanks for the fix. 👍

PR #909, opened 2023-02-01
fix: Proxy to defineProperty in ApiFormState

fenderil:
Change Proxy in ApiFormState to defineProperty's get descriptor since v7.13.0


bluebill1049:
thanks for the PR, would you like to fix the conflict?

bluebill1049:

PR #914, opened 2023-02-24
Fix: `cannot find IFormInput name` error in  code examples.

ambujsahu81:
### Description

- The ts code example in the section `Integrating with UI libraries`  in the get started page is missing the `IFormInput` interface.
- But the codeSandbox includes the `IFormInput` interface, so this pull request adds the same.
- It is being used  in the ts example code already, only the definition is missing.
https://github.com/react-hook-form/documentation/blob/a9d51013cd265cbbb7d40a53a6c4a09b97e04c67/src/components/codeExamples/getStarted.ts#L194

### DIFF
--- a/src/components/codeExamples/getStarted.ts
+++ b/src/components/codeExamples/getStarted.ts
@@ -182,8 +182,14 @@ export const uiLibraryHookInputTs = `import Select from "react-select";
 import { useForm, Controller, SubmitHandler } from "react-hook-form";
 import Input from "@material-ui/core/Input";
 
+interface IFormInput {
+  firstName: string;
+  lastName: string;
+  iceCreamType: { label: string; value: string };
+}
+
 const App = () => {
-  const { control, handleSubmit } = useForm({
+  const { control, handleSubmit } = useForm<IFormInput>({
     defaultValues: {
       firstName: '',
       lastName: '',
--- a/src/components/codeExamples/getStarted.ts
+++ b/src/components/codeExamples/getStarted.ts
@@ -189,7 +189,7 @@ interface IFormInput {
 }
 
 const App = () => {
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for the PR, it's actually not required when you have provided defaultValues, we can infer from that.

ambujsahu81:
> Thanks for the PR, it's actually not required when you have provided defaultValues, we can infer from that.

@bluebill1049  Thanks for reviewing 🙂  but when i use the code snippet that has the default values give me the following error.
 **Error :** Cannot find name 'IFormInput'.ts(2304) . This is because it is `IFormInput` already used in the code.

image

- Also in the `Integrating Controlled Inputs` next code snippet  it has default values and `IFormInput` interface defined too. 

**Example**
```ts
import { useForm, Controller, SubmitHandler } from "react-hook-form";
import { TextField, Checkbox } from "@material-ui/core";

// here also it is defined
interface IFormInputs {
  TextField: string
  MyCheckbox: boolean
}

function App() {
  const { handleSubmit, control, reset } = useForm<IFormInputs>({
    defaultValues: {
      MyCheckbox: false
    }
  });
  const onSubmit: SubmitHandler<IFormInputs> = data => console.log(data);
...
```ts

ambujsahu81:
@bluebill1049  will you please have another look at this one because it is fixing an error on ts code example. That code snippet already uses `IFormInputs` which is not defined . https://github.com/react-hook-form/documentation/blob/a9d51013cd265cbbb7d40a53a6c4a09b97e04c67/src/components/codeExamples/getStarted.ts#L194

bluebill1049:
aha thanks for fixing that, ideally we want to infer the type from `defaultValues`, but with your use case it's correct as well.

bluebill1049 (approved):

PR #918, opened 2023-03-09
add section for 3rd party bindings to resource page

aranlucas:
Adding section proposed in https://github.com/react-hook-form/documentation/issues/917

Copied contents of `Resources.json` to `resources.tsx` since that's where the edit page leads to.

Keeps same search API

Screenshot 2023-03-09 093247


bluebill1049 (approved):
LGTM!

PR #926, opened 2023-03-28
Resources: Use updated article versions

Clarity-89:
I've updated my articles from the resources section to use the version 7 of React Hook Form, so I've moved them into the v7 section. I've also added one other article about working with multi-step forms, which might be useful to the readers.


bluebill1049 (approved):
LGTM 👍

PR #929, opened 2023-04-02
fix: update useWatch hook documentation

### LINKED ISSUE #927: issue: Confusing useWatch documentation
### Version Number

7.43.8

### Codesandbox/Expo snack

https://codesandbox.io/s/headless-shadow-i6brkl?file=/src/App.js

### Steps to reproduce

1. Calling `useWatch` with a string (the field name) as an argument will return all form values.

### Expected behaviour

According to `useWatch` [documentation](https://react-hook-form.com/api/usewatch/#main), one could call `useWatch` passing a string as an argument and it would return the field's value.
image
This is not the case.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

RamonEspinosa:
# Description
This PR updates the documentation of the `useWatch` hook. The **Return**  section was confusing as it implied devs could pass a `string` as an argument to `useWatch` to get a field's value. Instead, they have to pass an object with a `name` property.

Closes #927 

### Before
image
### After
image

### DIFF
--- a/src/components/UseWatchContent.tsx
+++ b/src/components/UseWatchContent.tsx
@@ -137,7 +137,7 @@ export default function UseFieldArray({
             </tr>
             <tr>
               <td>
-                <code>useWatch('inputName')</code>
+                <code>useWatch({`{ name: 'inputName' }`})</code>
               </td>
               <td>
                 <code className={typographyStyles.typeText}>unknown</code>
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
LGTM! Thanks

PR #932, opened 2023-04-04
Replaced not existing style attribute for placeholder

KaanMol:
While looking at the documentation of React Hook Form for React Native, the example code gave an error because of a not existing style. 

I removed this and replaced it with a placeholder, now the code works directly.


bluebill1049:
thanks 🙏

PR #933, opened 2023-04-05
Fix/#930 shrinking input

### LINKED ISSUE #930: issue: Subscription checkboxes shrink when using uppercase
The subscription checkboxes shrink if uppercase is used.
RHF subscriptions

KaanMol:
When given a string that was too long, the input checkbox on the end would shrink. This has been fixed by giving it a min-width. On the mobile view the last char would get chopped of, for this reason also the char limit has been decreased by one.

closes #930


bluebill1049 (approved):
Thank you. LGTM

PR #942, opened 2023-04-17
Example in zod validation resolver doesn't work

AngyDev:
Hi,

in the zod resolver examples on the following page https://react-hook-form.com/api/useform/ it's not possible to run the code because the "age" field goes in error with the following message: `Expected number, received string`.
To resolve the error I added the `{ valueAsNumber: true }` in the age input.

There is the same error in the codesanbox examples, also in these examples, there is another error, the name type is a number instead a string. Is there a way to fix them? If you let me know I can do that.

Thanks


jorisre (approved):
Good catch thank you

PR #948, opened 2023-04-23
chore: remove default language from little-state-machine

aranlucas:
1. removes location from page props since these will not be available in NextJS. Since all of the location calls are on the client side (called within useEffect) they can be moved to window.location for access.
2. Removed defaultLanguage - This was confusing since little-state-machine is not available on server. The only way to do this with nextJS would have been to use the router locale, but that would probably mean integrating with more i18n things. For now, removing en until there's a better way to handle these translations more consistently.
3. Improved as many typescript errors as I could because nextJS is way more strict with type checking

Next one is NextJS actual migration (no more clean up).


aranlucas:
Sorry for all the changes 😅

bluebill1049:
np, reviewing now

bluebill1049 (approved):
Awesome! Thanks for simplifying this.

PR #950, opened 2023-04-23
feat: migrating to nextjs

aranlucas:
https://nextjs.org/docs/migrating/from-gatsby

Move all gatsby links to nextjs links
Removed all gatsby packages
Renamed api --> docs since it's a reserved namespace for nextjs
Added CSS variables rather than relying on "lightMode". https://github.com/pacocoursey/next-themes#avoid-hydration-mismatch . Probably where I will need guidance, as my CSS -foo isn't the best (but there are no hydration mismatches or flashes!)


aranlucas:
TODO: I don't think I understand Nav too well unfortunately. 

I need to find a way to separate Desktop Nav vs Mobile nav since I get them confused so easily

aranlucas:
I could make this even simpler by doing the rename from api -> docs while gatsby is still default. Might focus the dark theme issues more than now

aranlucas:
@bluebill1049 can you please approve the deployment? 

Fixed up the dark mode issues (that I saw) using global(".dark") and global(".light") CSS modules and I think it's looking pretty good.

Just want to verify from non-dev machine.

Thanks

bluebill1049:
works great with dark mode 👍

bluebill1049:
I will review it this week and merge it in.

aranlucas:
Let me know if there's anything I need to address. I see that I need to fix light mode select in API page.

bluebill1049 (approved):

PR #956, opened 2023-05-19
fix: dead link in mobile nav and footer

aranlucas:
Did another round of looking for dead links and found these two that were introduced when I separated resources


bluebill1049 (approved):

bluebill1049:
thanks 🙏

PR #957, opened 2023-05-19
fix: light background on mobile menu

aranlucas:
Light theme navbar has a global style none applied when it should only be done if screen size is 1024px +

image


bluebill1049 (commented):
Thanks!

PR #964, opened 2023-06-01
feat: moving FAQ to mdx content for easier editing

aranlucas:
As I mentioned before with NextJS migration, RHF can start taking advantage of MDX to write docs which should be simpler than having to write html.

In this PR:
* Migrating FAQ page to MDX. For now, this was the simplest page that I could migrate. They can eventually all start to be moved, but each component will take a while to add if there are custom styles (Headers have some custom styles)
* Created a new NewSideMenu component. Trying to follow the pattern in the react docs - https://github.com/reactjs/react.dev/blob/main/src/sidebarLearn.json. The idea is that we can't use refs as they are currently being used, so this needs to have a more "static" nature which should be fine.
* Introduced prism-react-renderer for prisma highlighting. I copied docusaurus in this department https://docusaurus.io/docs/markdown-features/code-blocks
* Changed the way code tabs work. Instead of having inline TS/JS used TabGroup Component. This was again inspired by docusaurus https://docusaurus.io/docs/markdown-features/code-blocks#multi-language-support-code-blocks
* Code blocks may have slightly different theme. Tried my best to copy current one. Here's what they look like now https://github.com/FormidableLabs/prism-react-renderer/blob/master/packages/prism-react-renderer/src/themes/dracula.ts


bluebill1049:
Thanks @aranlucas again! Did you receive a request for Paypal donation from open collective?

aranlucas:
Yes, I forgot about it after it asked for details. I'll fill it out now 👀

bluebill1049 (approved):

PR #969, opened 2023-06-04
fix: remove openbase link

aroyan:
Remove the Openbase feedback link in the footer since they have shut down their service
<br />
image


bluebill1049:
thanks, that's unfortunate.

PR #972, opened 2023-06-08
fix: adding redirect logic from api pages

aranlucas:
Adds a catch all api route for anything under "/api"

https://nextjs.org/docs/pages/building-your-application/routing/api-routes#optional-catch-all-api-routes

TODO: move to 301 in the future. Right now keeping at 307 since I had issues in testing if I messed up since it would have a wrong redirect.


bluebill1049 (approved):

bluebill1049:
LGTM

PR #975, opened 2023-06-14
fix: if docsearch is not loading, don't crash page

aranlucas:
Upgrading to https://docsearch.algolia.com/docs/DocSearch-v3 would be more ideal than this fix but it will prevent this script from crashing the page.

Tested by removing docsearch script from Head and loading the page.

Before, it would crash the page. Now, search is disabled instead.


bluebill1049 (approved):

bluebill1049:
Awesome thanks!

PR #982, opened 2023-06-23
fix: make start script reference dev script

Adrastopoulos:
The start script referenced a nonexistent `develop` script. Now updated to `dev` script.


bluebill1049 (approved):
thanks.

PR #986, opened 2023-06-26
fix: devtools dependency stripped out in prod

aranlucas:
## Problem
Oversight on my part.

Devtools was working in local development because process.env === "DEVELOPMENT"

When deploying to prod, DevTools became a no-op and does not render. https://github.com/react-hook-form/devtools/blob/master/rollup/writeCjsEntryFile.js

This changes the import path to use the development bundle and still appear in prod. 

This change got lost in trying to use the dynamic import as suggested by devtools repo for it to work in nextjs.

## Testing
* npm run dev - Worked as expected in development
* npm run build && npm run start - Build + start mimicks prod behaviour. Worked when importing the development bundle since it doesn't become a no-op

### DIFF
--- a/src/components/DevTools.tsx
+++ b/src/components/DevTools.tsx
@@ -15,9 +15,13 @@ import getStartedStyle from "./GetStarted.module.css"
 import styles from "./DevTools.module.css"
 import ClipBoard from "./ClipBoard"
 import dynamic from "next/dynamic"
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks for the fix 🙏

aranlucas:
Does it count if it was my fault? haha

PR #990, opened 2023-07-06
feat: adding minimal tailwind setup

aranlucas:
https://stackoverflow.com/questions/71783177/remove-specific-style-from-tailwind-base

The idea would be to start removing some of the global styling in layout.css and creating components using tailwind rather than CSS modules https://ui.shadcn.com/docs/components/

This can be done when when moving to app directory or as a slow migration.

Note: not everyone likes tailwind, so leaving this as optional but it would help to rewrite some of the components using radix + tailwind for accessibility reasons.


aranlucas:
First step toward https://github.com/react-hook-form/documentation/issues/915#issuecomment-1597798175

PR #991, opened 2023-07-06
feat: adding form and getFieldState to mdx

aranlucas:
Adding getFieldState + form MDX files

I'm still not super happy with the inline code but this will unblock some moving tables with code.

Missing 6 more pages to migrate to MDX
* formstate
* register
* usecontroller

* usefieldarray
* useformstate


bluebill1049 (approved):

bluebill1049:
Screenshot 2023-07-08 at 6 34 22 pm
crashed

bluebill1049 (changes requested):
route is broken.

aranlucas:
The root cause is the sidebar is not being resolved. I'm interested in why it's happening now since this change doesn't seem to have anything that could affect that, but I have a local change to fix it

aranlucas:
There seems to have been 2 issues.

* SideMenu showing up null
* git file naming casing. I changed it locally to be lowercase and it worked as expected, but git doesn't detect those types of changes 😞 

```javascript
  const doc = allDocs.find(
    (doc) => doc._raw.flattenedPath === params?.slug?.join("/")
  )
```

bluebill1049 (approved):
Awesome thanks 👍

PR #1018, opened 2023-10-12
Fix React Native documentation 

fredericlesueurs:
Hi,

Fix React Native documentation, because if you write `onChangeText={onChange}` your input was laggy.
The correct code was `onChangeText={(value) => onChange(value)}`.


bluebill1049:
hi,

thanks for the PR.

> Fix React Native documentation, because if you write onChangeText={onChange} your input was laggy.

are you sure this is the issue?

PR #1056, opened 2024-03-29
Improve `watch`  to reflect overloads

### LINKED ISSUE #1040: Document watch callback type parameter
**Is your feature request related to a problem? Please describe.**
Using the `watch` API in `useEffect` like so:
```javascript
useEffect(() => {
  watch((data, {name, type}) => {
    console.log(data, name, type)
  })
})
```

It seems `type` indicates who initiated the update, because a user change results in `type: 'change'`, while a `setValue` call results in `type: undefined`. I'm actually using it avoid endless watch calls as I call `setValue` inside `watch`, which results in a new `watch` callback call. I'd like to ask for documentation of this `type` field.

**Describe the solution you'd like**
Document the possible options for the `type` attribute in `watch` callback.

**Describe alternatives you've considered**
I can probably hunt down all usages within the code but it seems suboptimal. The current documentation just mentions the presence of this field and it's even typed as just `string`, which doesn't convey much information about it.

**Additional context**
_React Hook Form seems like a really neat and advanced library - well done!_

honey32:
relates to #1040

I tried to make description in https://react-hook-form.com/docs/useform/watch to reflect _real_ overload of the function according to [its JSDoc](https://github.com/react-hook-form/react-hook-form/blob/306627e385738e322cad8dff9e752678f8210020/src/types/form.ts#L368)

I'm not sure about description, but I'm sure it would be better if organized like this, because it has _two different usages_, which,   without explanation, may lead beginners to confusion.

I'm willing to make some additional edit if needed to make description more precise.


bluebill1049 (approved):
Nice! Ty

bluebill1049 (approved):

PR #1064, opened 2024-04-28
Add note to Controller documentation - FE-1221

mbrowne:
I was implementing a feature recently where I needed to control the value of a dropdown menu via a state variable that existed outside the form. I eventually realized the most efficient way to do this was to use the `values` option of `useForm`. This PR adds a note to hopefully make that clearer for others.


bluebill1049 (approved):

bluebill1049:

Issue #1083, opened 2024-08-29
Side Menu doesn't change bg color when Documentation Scroll 


Srabon444:
As I am an advocate for user friendliness; I have to mention that when the user goes through the documentation, the side menu doesn't get highlighted. It causes the user to wonder which part he/she is in. It would be a great idea if you just implemented this tiny feature.

Thanks for your kind and hard work.

Tiru-99:
Hey @bluebill1049 and @Srabon444 , I would love to contribute to the project by working on this issue. If it’s possible, could you please assign it to me and add the Hacktoberfest tag?

unclebay143:
This issue can be closed now, it has been addressed here: https://github.com/react-hook-form/documentation/pull/1085
@bluebill1049

bluebill1049:
ty 👍

PR #1085, opened 2024-10-10
Added scroll based menu highlight feature

Tiru-99:
# Enhanced Side Menu with Active Section Tracking #1083 
## Overview
This PR implements an enhanced side menu component that automatically highlights the current section as users scroll through the page. The menu now provides visual feedback for navigation, improving the overall user experience in our single-page documentation layout.

## Key Changes
- Implemented scroll-based active section detection
- Added visual highlighting for both parent and child menu items
- Introduced smooth scrolling behavior when clicking menu items
- Updated styling to support the new interactive features

## Implementation Details
- Uses `useEffect` and scroll event listener to track the current section
- Calculates section positions dynamically to handle content updates
- Implements a `100px` offset for better UX when transitioning between sections
- Handles both top-level and nested menu items

##Execution Video 

https://github.com/user-attachments/assets/707c5a0c-22be-4682-a697-1ccae702cf24

### Code Example

```typescript
// Function to determine which section is currently in view
const handleScroll = () => {
//Create an array of all sections with their IDs and positions
  const sections = pages.flatMap(page => 
    [page, ...(page.pages || [])].map(p => ({
      id: p.pathname.replace('#', ''),
      top: document.getElementById(p.pathname.replace('#', ''))?.offsetTop || 0
    }))
  )

//Find the section that's currently in view
  const currentSection = sections.reduce((acc, section) => {
    const { id, top } = section
    if (window.scrollY >= top - 100) {   // 100px offset for better UX
      return id
    }
    return acc
  }, '')

 //Update the state with the current section
  setActiveSection(currentSection)
}

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -1,13 +1,50 @@
+import { useEffect, useState } from "react"
 import Link from "next/link"
-import colors from "../../styles/colors"
 import styles from "./SideMenu.module.css"
 import typographyStyles from "../../styles/typography.module.css"
-import { useRouter } from "next/router"
+import colors from "../../styles/colors"
 import { Pages } from "../../types/types"
 
 function Menu({ pages = [] }: { pages: Pages }) {
-  const router = useRouter()
-  const { asPath: pathname } = router
+  const [activeSection, setActiveSection] = useState<string>("")
+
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
Thanks

PR #1097, opened 2024-12-13
feat(useForm): add “`useForm` return and `useEffect` dependencies”

marcalexiei:
Related to https://github.com/react-hook-form/react-hook-form/issues/12463.

Adds a new section in useForm page informing about the future problems of adding the whole `useForm` return to the dependencies list of a `useEffect`


bluebill1049:
Amazing thank you!

Issue #1098, opened 2024-12-13
Repo maintenance improvements


marcalexiei:
When opening #1097 I noticed a few things that might be fixed / improved about repository maintenance

## 1. Double lock files

The project contains both `pnpm-lock.yml` and `yarn.lock`. 

### Proposed solution

`pnpm-lock.yml` was update more recently so I would remove `yarn.lock` from version control and update all `yarn` command across the repository using `pnpm`

## 2. `format` script is not valid

Running `pnpm format` produce an error because the prettier command is missing the `[file/dir/glob ...]`.

### Proposed solution

Update the script adding a `.` and add not relevant folders / files to `.prettierignore`

## 3. `pnpm run lint` returns an error

Running `pnpm run lint` produce many errors / warnings. 

<details>
<summary>`pnpm run lint` Output</summary>

```text
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.2.0

YOUR TYPESCRIPT VERSION: 5.6.3

Please only submit bug reports when using the officially supported version.

=============

./src/components/ApiFormState.tsx
9:39  Warning: Unexpected any. Specify a different type.  @typescript-eslint/no-explicit-any

./src/components/ApiRefTable.tsx
10:53  Warning: Unexpected any. Specify a different type.  @typescript-eslint/no-explicit-any

./src/components/BeekaiBuilderPage.tsx
13:11  Warning: Using `<img>` could result in slower LCP and higher bandwidth. Consider using `<Image />` from `next/image` to automatically optimize images. This may incur additional usage or cost from your provider. See: https://nextjs.org/docs/messages/no-img-element  @next/next/no-img-element

./src/components/BuilderPage.tsx
113:6  Warning: React Hook useEffect has a missing dependency: 'setValue'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps
117:6  Warning: React Hook useEffect has a missing dependency: 'setValue'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps
121:6  Warning: React Hook useEffect has missing dependencies: 'editFormData.required' and 'setValue'. Either include them or remove the dependency array.  react-hooks/exhaustive-deps

./src/components/CodeCompareSection.tsx
5:8  Warning: 'colors' is defined but never used.  @typescript-eslint/no-unused-vars

[...]

PR #1105, opened 2024-12-17
feat(deps): update next to v15.1

marcalexiei:
Part of #1098.

I changed the order of the tasks.
Getting `eslint@9` will be easier if we are using `next@15`.

I updated all next js dependencies using the [official codemod](https://nextjs.org/docs/app/building-your-application/upgrading/version-15).

When trying out the site I encountered a few non-blocking errors related to `key` prop management
in the following files:
- `src/components/FormFields.tsx` (duplicate / missing `key` prop)
- `src/components/mdx/code.tsx` (`key` prop should not be passed via spread operator)

----

I navigated a few pages and test the form builder running the site via `pnpm run build` and then `pnpm run start`.

> [!NOTE]
> React has been update to `v18.3.1`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,6 @@ module.exports = {
   },
   extends: [
     "plugin:react/recommended",
-    "plugin:react-hooks/recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:jsx-a11y/recommended",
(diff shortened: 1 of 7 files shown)


bluebill1049 (approved):
Nice! Thank you!

PR #1106, opened 2024-12-20
chore(deps-dev): update eslint to v9.17

marcalexiei:
Part of #1098

Migrated eslint to v9.17. 
All eslint plugins have also been updated to their latest version.

Additional unrelated changes:
- `@types/node` have been moved inside `devDependencies`
- `date-fns` dependency is not used so I have removed it


bluebill1049 (approved):
LGTM

PR #1132, opened 2025-03-20
Warn users that resetting the form while passing values will also reset defaultValues

shaundavin13:
Added documentation that warns users that when calling `reset` with `values`, the form's `defaultValues` updates accordingly, which means subsequent calls to `reset` will use the previously passed `values` instead of the form's original `defaultValues`. This helps developers who call `reset` with API responses to be aware that subsequent `reset` only revert back to that API response


bluebill1049 (approved):

bluebill1049:

PR #1134, opened 2025-04-10
fix(register): hard to read useForm/register option section in light mode due to dark background

JjungminLee:
Hi! I noticed that the options section of useForm/register was hard to read in light mode due to the dark background (https://react-hook-form.com/docs/useform/register).
So, I added new CSS properties to improve its visibility!
###  Before
- light
image
- dark
image

### After

- light
image
- dark
image

### 📣 Notice
Since I ran pnpm run format:fix, some unrelated file changes were included in the commit.


bluebill1049 (approved):
nice thanks

PR #1174, opened 2025-10-19
Fix: Copy button hidden on small screens (<786px)

LuckySilver0021:
Copy button was not showing up below screen width of 768px, fixed it.

### DIFF
--- a/src/content/get-started.mdx
+++ b/src/content/get-started.mdx
@@ -14,7 +14,7 @@ npm install react-hook-form
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


LuckySilver0021 (commented):
its showing more files changed that actually changed because accidently I changed syntax of it , but the core change reamins only in the : src/components/CodeArea.module.css file
lmk if its fine  @bluebill1049

LuckySilver0021 (commented):
.

bluebill1049:
let's not do that 
image
thanks for the PR tho.

PR #1214, opened 2026-05-26
docs: correct setValues callback return type to TFieldValues

dfedoryshchev:
The `setValues` callback return type was documented as `(data: TFieldValues) => Partial<TFieldValues>`, but the source type `UseFormSetValues` declares it as `(formValues: TFieldValues) => TFieldValues`.

Updated the heading signature, the Props table cell, and the Rules admonition wording to match the source. The existing callback example already returns `{ ...data, name: "test" }` (full `TFieldValues`), which is now consistent with the documented type.


bluebill1049 (approved):
lgtm

PR #1230, opened 2026-08-13
fix sticky sidebar

Kunal-0117:
## Summary

  This fixes the sidebar not staying sticky on pages outside `/docs/*`.

  The page animation was keeping a transform active after it finished, which
  interfered with the sidebar’s fixed positioning. Removing the animation fill mode
  keeps the entrance animation smooth while allowing the sidebar to remain sticky.

  ## Verification

  - Tested the sidebar while scrolling
  - Prettier passed
  - TypeScript passed
  - Production build passed

### DIFF
--- a/src/components/layout.css
+++ b/src/components/layout.css
@@ -848,7 +848,7 @@ pre[class*="language-"] {
 }
 
 .pageTransition {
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
LGTM

PR #1231, opened 2026-08-29
docs: document exact ancestor-path matching since 7.81.0

### LINKED ISSUE #13550: issue: Clearing a value that is an object does not remove the value from useController
### Version Number

7.79.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/yrk669

### Steps to reproduce

1. click on "clear" button
2. the "plain" input gets cleared
3. the object one "data.type" does not

### Expected behaviour

When the value is an object and I set it to null, the useControllers that are subscribed to its properties to be null too. Currently the form values are out of sync with the value from useController. 
The values from useWatch and useController are different

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

22mb:
## Summary

Since v7.81.0, a subscription with `exact: true` also fires when an **ancestor path** of the subscribed name is updated. Before 7.81.0, `exact: true` only matched the identical name. The reverse direction is unchanged: with `exact: true`, a change to a nested child path still does not notify a subscriber of the parent name (see react-hook-form/react-hook-form#12867).

The behavior changed in react-hook-form/react-hook-form#13553 (fixes react-hook-form/react-hook-form#13550, first released in v7.81.0), but neither the release notes nor the CHANGELOG mention `exact`, and the docs still describe `exact: true` as "an exact match for input name subscriptions". I ran into this when upgrading.

Source of truth (`src/logic/shouldSubscribeByName.ts`):

```diff
 (exact
-  ? currentName === signalName
+  ? currentName === signalName || currentName.startsWith(signalName + '.')
   : currentName.startsWith(signalName) ||
     signalName.startsWith(currentName)),
```

For a subscription to `"users.0.name"`:

| Change to                          | `exact: false` | `exact: true` (< 7.81.0) | `exact: true` (>= 7.81.0) |
| ---------------------------------- | -------------- | ------------------------ | ------------------------- |
| `"users.0.name"`                   | fires          | fires                    | fires                     |
| `"users.0"` / `"users"` (ancestor) | fires          | does not fire            | **fires**                 |

For a subscription to `"users"`, a change to `"users.0.name"` (descendant) fires only with `exact: false` — unchanged.

## Changes

Updated the `exact` row in every props table that has one, following the existing `<Since version="..." />` convention (as in #1206, #1216, #1219):

- `useWatch` (`usewatch.mdx`), `useFormState` (`useformstate.mdx`), `subscribe` (`useform/subscribe.mdx`): describe both `false` and `true`, including the 7.81.0 ancestor-path behavior and the asymmetry.
- `useController` (`usecontroller.mdx`), `Controller` (`usecontroller/controller.mdx`): same, phrased for the controlled field (default `exact: true`), with a hint that `exact: false` also receives nested updates.
- `Watch` (`usewatch/watch.mdx`), `FormState` (`useformstate/formstate.mdx`): short version linking to the hook docs.
- `subscribe`'s `exact` type now shows its default (`boolean = false`), consistent with the other tables (`exact` is passed through as-is, so it is falsy unless set).

The rest of the diff is prettier re-padding the tables; `pnpm format` passes.


bluebill1049 (approved):
LGTM

